Issue #1, opened 2021-12-17
-G flag should set the environment variables GROUP and JOBSUB_GROUP


shreyb:
Currently, the -G flag to jobsub_lite commands sets the group of the submission, which is only used in tarfile upload.  We need to also populate the GROUP environment variable with this value for submission to the DUNE global pool, since otherwise, token generation (and thus authorization to submit) fails.

The current set of steps needed to submit a job to GPGrid is:

```source /grid/fermiapp/products/common/etc/setups
export GROUP=dune
jobsub_submit -G dune  --resource-provides=usage_model=OPPORTUNISTIC,DEDICATED  file:///home/sbhat/TestJobs/basicsleep.sh 123
```
For the DUNE pool, the process is:

```source /grid/fermiapp/products/common/etc/setups
export _condor_COLLECTOR_HOST=dunegpcoll02.fnal.gov
export GROUP=dune
jobsub_submit -G dune  --resource-provides=usage_model=OPPORTUNISTIC,DEDICATED  file:///home/sbhat/TestJobs/basicsleep.sh 123
```
The third line is unnecessary, since we're already specifying group.

marcmengel:
So I believe the pull request resolves this.  Closing it out

PR #3, opened 2021-12-17
Store -G/--group value into environment variable GROUP

shreyb:
Added a callback `argparse.Action` to get_parser so that any inheriting module will automatically store the value of args.group into the environment variable GROUP.

For the shell scripts (jobsub_cmd, jobsub_q), pull the group flag/value out of the args, store it in GROUP, and modify args so that condor doesn't complain about the -G/--group flag getting passed.

### DIFF
--- a/bin/jobsub_cmd
+++ b/bin/jobsub_cmd
@@ -28,6 +28,20 @@ then
     args="-name $schedd $args"
 fi
 
+# Check group and pull it into environment
+case "x$*" in
+*--group*)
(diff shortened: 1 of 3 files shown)


shreyb:
I tested this with jobsub_submit and jobsub_q on both jobsubdevgpvm01 and DUNE pool.  I think I got the `sed` expression right, but please double check that for me!

shreyb:
Oops - forgot to say that this addresses #1

marcmengel:
Looks good, Thanks!

Issue #13, opened 2022-02-09
--role value is not passed to templates, resulting in submission failures


shreyb:
When I try to submit a job using the production role, the following happens (cut some of output here):

```
jobsub_submit -G dune --role=production   --resource-provides=usage_model=DEDICATED,OPPORTUNISTIC,OFFSITE --devserver file:///usr/bin/printenv
...
Running: _condor_CREDD_HOST=jobsubdevgpvm01.fnal.gov BEARER_TOKEN_FILE=/tmp/bt_token_dune_Analysis_10610 /usr/bin/condor_submit -spool -pool gpcollector04.fnal.gov -remote jobsubdevgpvm01.fnal.gov /home/sbhat/.jobsub_lite/dune/sbhat/2022_02_09_135318.3792bd86-4bf1-46b3-a639-09114db72cb4/simple.cmd
...
Checking if -s dune credentials exist
Attempting to get tokens for dune
Credkey from /home/sbhat/.config/htgettoken/credkey-dune-default: sbhat
Attempting to get bearer token from https://fermicloud543.fnal.gov:8200
  using vault token from /tmp/vt_u10610-dune
  at path secret/oauth/creds/dune/sbhat:default
Storing bearer token in /tmp/bt_token_dune_Analysis_10610
Submitting job(s).
1 job(s) submitted to cluster 40921@jobsubdevgpvm01.fnal.gov
Output will be in /home/sbhat/.jobsub_lite/dune/sbhat/2022_02_09_135318.3792bd86-4bf1-46b3-a639-09114db72cb4 after running jobsub_transfer_data.
```

So the submission works, but note that the bearer token filename has "Analysis" in there.  Looking at the bearer token itself (again omitting some lines):

```
$ httokendecode /tmp/bt_token_dune_Analysis_10610
{
  "sub": "sbhat@fnal.gov",
  "nbf": 1644436373,
  "scope": "storage.create:/dune/scratch/users/sbhat compute.create compute.read compute.cancel compute.modify storage.read:/dune",
  "wlcg.groups": [
    "/dune"
  ],

}
```

the correct `/dune/production` group hasn't been included in the wlcg.groups entry.  This is because the line here (and corresponding ones in other templates)

https://github.com/marcmengel/jobsub_lite/blob/9a25dfa5711fafc547274467111398c1afcec9ee/templates/simple/simple.cmd#L62

[...]

PR #14, opened 2022-02-11
Added role to templates for JDFs that jobsub_submit generates

shreyb:
Add role to template so that if we specify --role on the command line, that gets passed to condor JDF.


shreyb:
Addresses issue #13

goodenou (approved):
This looks good to me.

PR #15, opened 2022-02-23
Changes to spec file to make it work with a vanilla development env

shreyb:
There were a few remnants in the spec file (__pycache__ files, byte-code files) that would only work if source code was run.  This change is to allow for building of the RPM in the case that tests were run on a different machine, and strips out the installation and inclusion of cached and byte-compiled files.


retzkek (approved):
LGTM

Issue #30, opened 2022-03-23
When a non-Analysis role is given, make sure we set BEARER_TOKEN_FILE (and thus BEARER_TOKEN) correctly in wrapper scripts


shreyb:
When a non-Analysis role is given, make sure we set BEARER_TOKEN_FILE (and thus BEARER_TOKEN) correctly in wrapper scripts.  For example, for role production, experiment dune, the jobsub_submit line will be something like:

```
jobsub_submit -G dune --role production <args> file:///<path>/<to>/<executable>
```

In this case, in the wrapper scripts, (like https://github.com/marcmengel/jobsub_lite/blob/master/templates/simple/simple.sh#L10), we need to make sure that not only is the group templated, but the role.  The convention is that the bearer token file for a production token will be named something like dune_production.use.  So in general, <group>_<role>.use.  

This should be pretty similar to the changes that had to be made for #13 .

PR #32, opened 2022-03-28
Update wrapper script templates to support roles

shreyb:
The wrapper scripts were not correctly using the value of the --role flag to modify the default paths for the environment variables `BEARER_TOKEN_FILE` and `BEARER_TOKEN`.  This PR fixes that (#30 )

### DIFF
--- a/templates/dag/dagman_wrapper.sh
+++ b/templates/dag/dagman_wrapper.sh
@@ -3,6 +3,11 @@
 # condor wants to copy in the condor_dagman executable, but we
 # want it to run the local one, so we give it this one...
 
-export BEARER_TOKEN_FILE=$_CONDOR_CREDS/{{group}}.use
+{% if role and role != 'Analysis' %}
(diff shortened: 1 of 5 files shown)


shreyb:
This has been tested on fifeutilgpvm01 with a dunepro job (133@dunegpschedd02.fnal.gov).

templates/dag/dagman_wrapper.sh
```diff
@@ -3,6 +3,11 @@
 # condor wants to copy in the condor_dagman executable, but we
 # want it to run the local one, so we give it this one...
 
+{% if role and role != 'Analysis' %}
```
retzkek, line 6: does the dagman actually need (or even get) the token?

retzkek: Sure that works for me

retzkek (approved):
LGTM

PR #35, opened 2022-04-13
don't set BEARER_TOKEN in wrappers

### LINKED ISSUE #33: Move setting of BEARER_TOKEN into ifdh and remove from jobsub_lite
Move setting of BEARER_TOKEN into ifdh and remove from jobsub_lite.  This relies on ifdh setting BEARER_TOKEN in the environment by reading from BEARER_TOKEN_FILE.  This change should mostly need to take place in the wrapper scripts. (see #30)

retzkek:
Setting `BEARER_TOKEN` in the job environment is a potential security leak, exposing tokens in log files, etc. by the common practice to print the jobs environment for debugging. It's now set internally by ifdh v2_6_3 when a token file is discovered. Eventually gfal2 will support token discovery: https://its.cern.ch/jira/browse/DMC-1226.

Fixes #33

### DIFF
--- a/templates/dag/dagman_wrapper.sh
+++ b/templates/dag/dagman_wrapper.sh
@@ -8,6 +8,5 @@ export BEARER_TOKEN_FILE=$_CONDOR_CREDS/{{group}}_{{role}}.use
 {% else %}
 export BEARER_TOKEN_FILE=$_CONDOR_CREDS/{{group}}.use
 {% endif %}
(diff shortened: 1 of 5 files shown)


shreyb:
This looks fine, but is ifdh 2_6_3 out yet?   I believe Ken has a few tests where he sets up a specific ifdh release (rather than pulling the latest), so we'll have to let him know.

retzkek:
It's "released" in that it's in common products and declared "test". It's not in scisoft or larsoft yet, which is probably where Ken needs it.

PR #36, opened 2022-04-21
Issue12

goodenou:
test

### DIFF
--- a/bin/condor_submit
+++ b/bin/condor_submit
@@ -68,7 +68,9 @@ def main():
     """
     cmd_args = sys.argv[1:]
 
-    proxy, token = get_creds(args)
+#   these args may/may not come from the input args
+    cred_args = {}
+    proxy, token = get_creds(cred_args)
 
     # put in environment for condor libs to use
     os.environ['X509_USER_PROXY'] = proxy
--- a/bin/condor_submit
+++ b/bin/condor_submit
@@ -22,14 +22,8 @@ sys.path = [ x for x in sys.path if x[:6] == '/usr/l' ]
 
 import os
 import os.path
-import re
-import glob
-import requests
-import subprocess
-import datetime
-import uuid
-import socket
-import shutil
+import argparse
+import htcondor
 
 #
 # we are in prefix/bin/jobsub_submit, so find our prefix
@@ -45,16 +39,23 @@ if PREFIX == '/usr':
 # this means we don't need fancy ups dependencies..
 #
 sys.path.append(os.path.join(PREFIX,"lib"))
+import packages
 
 #
 # import our local parts
 #
-from get_parser import get_parser 
-from condor import  get_schedd, submit, submit_dag
-from dagnabbit import parse_dagnabbit
+from condor import get_schedd
 from creds import get_creds
-from tarfiles import do_tarballs
-from utils import fixquote, set_extras_n_fix_units
+
+COLLECTOR_HOST = htcondor.param.get("COLLECTOR_HOST", "gpcollector03.fnal.gov")
+
+class StoreGroupinEnvironment(argparse.Action):
+    """Action to store the given group in the GROUP environment variable"""
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        os.environ["GROUP"] = values
+        setattr(namespace, self.dest, values)
+
 
 
 def main():
@@ -63,23 +64,101 @@ def main():
         - get credentials
         - handle tarfile options
         - set added values from environment, etc.
-        - convert/render template files to submission files
         - launch
     """
-    cmd_args = sys.argv[1:]
 
-#   these args may/may not come from the input args
+    condor_parser = argparse.ArgumentParser(
+        description='Queue jobs for execution under HTCondor without the jobsub wrapper.')
+
+    condor_parser.add_argument(
+        "-G",
+        "--group",
+        required=False,
+        help="Group/Experiment/Subgroup for priorities and accounting",
+        action=StoreGroupinEnvironment)
+
+    condor_parser.add_argument(
+        "-name",
+        required=False,
+        help="Submit to the specified condor_schedd")
+
+    condor_parser.add_argument(
+        "-spool",
+        required=False,
+        help="Spool all required input files, job event log, and proxy over the connection to the condor_schedd",
+        action='store_true')
+
+    condor_parser.add_argument(
+        "-remote",
+        required=False,
+        help="Submit to the specified condor_schedd, spooling all required input files over the network connection")
+
+    condor_parser.add_argument(
+        "--credentials",
+        help="credential information")
+
+
+#   PARSE ARGS TO GET SUBMISSION ARGUMENTS
+
+    args_tuple = condor_parser.parse_known_args()
+
+    if args_tuple[0].group is not None:
+        group = args_tuple[0].group
+    else:
+        group = os.environ["GROUP"]
+
+    if args_tuple[0].name is not None:
+        schedd_name = args_tuple[0].name
+    elif args_tuple[0].remote is not None:
+        schedd_name = args_tuple[0].remote
+    else:
+        schedd_add = get_schedd({"group": group})
+        schedd_name = schedd_add.eval("Machine")
+    
+    cmd_args = args_tuple[1]
+
+    print("schedd_name is : %s" % schedd_name)
+    print("cmd_args is : %s" % cmd_args)
+
+
+    if args_tuple[0].spool:
+       schedd_args = "-name '%s' -spool" %(schedd_name)
+    else:
+       schedd_args = "-remote '%s'" %(schedd_name)
+
+
+#   GET CREDENTIALS
+#   these args may/may not come from the input args. for now let's keep it empty  
     cred_args = {}
-    proxy, token = get_creds(cred_args)
+    if args_tuple[0].credentials is not None:
+        cred_args = args_tuple[0].credentials
 
-    # put in environment for condor libs to use
-    os.environ['X509_USER_PROXY'] = proxy
-    os.environ['BEARER_TOKEN_FILE'] = token
+    proxy, token = get_creds(cred_args)   
+    print("proxy is : %s" % proxy)
+    print("token is : %s" % token)
 
-    schedd_add = get_schedd({})
-    schedd_name = schedd_add.eval("Machine")
+# put in environment for condor libs to use
+    os.environ["X509_USER_PROXY"] = proxy
+    os.environ["BEARER_TOKEN_FILE"] = token
 
-    submit('', varg, schedd_add, cmd_args)
+
+#   CREATE SUBMIT COMMAND
+    cmd = "/usr/bin/condor_submit -pool %s %s %s" % (
+            COLLECTOR_HOST,
+            schedd_args,
+            " ".join(["'%s'" % x for x in cmd_args]),
+        )
+    cmd = "BEARER_TOKEN_FILE=%s %s" % (os.environ["BEARER_TOKEN_FILE"], cmd)
+    cmd = "_condor_CREDD_HOST=%s %s" % (schedd_name, cmd)
+    packages.orig_env()
+    print("Running: %s" % cmd)
+
+    cmd = '%s | sed -e "s/\\(submitted.to.cluster.*\\)\\./\\1@%s/"' % (
+        cmd,
+        schedd_name,
+    )
+    os.system(cmd)
+
 
 if __name__ == "__main__":
     main()
--- a/bin/jobsub_submit
+++ b/bin/jobsub_submit
@@ -162,7 +162,7 @@ def main():
     args = parser.parse_args()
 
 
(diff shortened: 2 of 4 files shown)


bin/condor_submit
```diff
@@ -63,21 +64,101 @@ def main():
...
+
+# put in environment for condor libs to use
+    os.environ["X509_USER_PROXY"] = proxy
+    os.environ["BEARER_TOKEN_FILE"] = token
+
 
-    # put in environment for condor libs to use
-    os.environ['X509_USER_PROXY'] = proxy
-    os.environ['BEARER_TOKEN_FILE'] = token
+#   CREATE SUBMIT COMMAND
```
shreyb, line 145: Is there any reason we need to replicate the code from `condor.py:submit()` here rather than using the `submit` call already present?  I'm loathe to have two different ways we submit jobs.

Would it be possible to, for example, pass the information you've parsed using vargs to `submit`, and then make hopefully slight modifications to the `submit` function itself to support the flags you've added support for here?  Let me know if that doesn't make sense.

bin/condor_submit
```diff
@@ -63,21 +64,101 @@ def main():
...
+        action='store_true')
+
+    condor_parser.add_argument(
+        "-remote",
+        required=False,
+        help="Submit to the specified condor_schedd, spooling all required input files over the network connection")
+
+    condor_parser.add_argument(
+        "--credentials",
+        help="credential information")
```
shreyb, line 98: Currently this doesn't support anything, right?  Do you think it's best to comment this out for now, or perhaps add a note saying that it's not implemented yet?

lib/condor.py
```diff
@@ -95,20 +95,24 @@ def load_submit_file(filename):
...
+
+    print("cmd=args: %s" % cmd_args)
+
     schedd_name = schedd_add.eval("Machine")
     schedd = htcondor.Schedd(schedd_add)
     print("schedd: %s" % schedd_name)
 
     if True:
-        cmd = "/usr/bin/condor_submit -spool -pool %s -remote %s %s %s" % (
+        cmd = "/usr/bin/condor_submit -pool %s -remote %s %s %s" % (
```
shreyb, line 115: Why is this change needed?  If I'm looking at this correctly, it'll change current jobsub_submit as well, which we shouldn't do unless we know it's necessary.

lib/creds.py
```diff
@@ -14,19 +14,23 @@
...
     myproxy, nor does it yet deal with tokens, but those should
     be done here as needed.
     """
-    f = os.popen("fake_ifdh getProxy","r") 
+    f = subprocess.Popen('fake_ifdh getProxy', shell=True, bufsize=0, stdout=subprocess.PIPE, universal_newlines=True).stdout
     p = f.read().strip()
     f.close()
-    f = os.popen("fake_ifdh getToken","r") 
+
+    f = subprocess.Popen('fake_ifdh getToken', shell=True, bufsize=0, stdout=subprocess.PIPE, universal_newlines=True).stdout
```
shreyb, line 29: Same as line 25.

shreyb (changes requested):
This is very thorough and well-thought out.  My big concern is that we're replicating `condor.py:submit()` code in `condor_submit`, which seems to be unnecessary.  Please see my comments in `condor_submit`, and let me know if you have any questions.  Thanks very much!

bin/condor_submit
```diff
@@ -63,21 +64,101 @@ def main():
...
+
+
+#   GET CREDENTIALS
+#   these args may/may not come from the input args. for now let's keep it empty  
+    cred_args = {}
+    if args_tuple[0].credentials is not None:
+        cred_args = args_tuple[0].credentials
+
+    proxy, token = get_creds(cred_args)   
+    print("proxy is : %s" % proxy)
```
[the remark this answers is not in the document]
goodenou: Sure, or we can just get rid of the print statements. Do you have a preference?

bin/condor_submit
goodenou: I will look into it. I felt I was going to have a bunch of "if condor_submit, do this, else if jobsub_submit do this other thing". Is that ok?

bin/condor_submit
goodenou: Correct. I will comment that it is not implemented.

lib/creds.py
```diff
@@ -14,19 +14,23 @@
...
+import subprocess
 
 
 def get_creds(args={}):
     """get credentials -- Note this does not currently push to
     myproxy, nor does it yet deal with tokens, but those should
     be done here as needed.
     """
-    f = os.popen("fake_ifdh getProxy","r") 
+    f = subprocess.Popen('fake_ifdh getProxy', shell=True, bufsize=0, stdout=subprocess.PIPE, universal_newlines=True).stdout
```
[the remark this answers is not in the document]
goodenou: I can make that change. I was not aware of suppress.run.

bin/condor_submit
```diff
@@ -63,21 +64,101 @@ def main():
...
+        help="Submit to the specified condor_schedd, spooling all required input files over the network connection")
+
+    condor_parser.add_argument(
+        "--credentials",
+        help="credential information")
+
+
+#   PARSE ARGS TO GET SUBMISSION ARGUMENTS
+
+    args_tuple = condor_parser.parse_known_args()
```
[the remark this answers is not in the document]
goodenou: Will look into this.

bin/jobsub_submit
```diff
@@ -162,7 +162,7 @@ def main():
     args = parser.parse_args()
 
 
-    proxy, token = get_creds(args)
+    proxy, token = get_creds()
```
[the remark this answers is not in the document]
goodenou: The get_creds function can take args, but in this case, it doesn't need them. If they are not passed, then an empty dict is assumed.

bin/condor_submit
shreyb: How many of these kinds of conditionals would you think you need to add to `condor.py:submit()`?  I'd hope it's only a couple of differences, if any, between the two.  And since the signature for that `submit()` function only cares about sets of args (`vargs`, `cmd_args`), I'm hoping not too much will change.  Please let me know if that's not the case.

lib/creds.py
goodenou: Change made. Update to PR forthcoming.

lib/creds.py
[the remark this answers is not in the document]
goodenou: Change made. Update to PR forthcoming.

bin/condor_submit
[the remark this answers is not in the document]
goodenou: Change made. Update to PR forthcoming.

bin/condor_submit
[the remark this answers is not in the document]
goodenou: Change made. Update to PR forthcoming.

bin/condor_submit
goodenou: I made the changes. I like the result.

lib/creds.py
```diff
@@ -14,19 +14,21 @@
...
     """
-    f = os.popen("fake_ifdh getProxy","r") 
-    p = f.read().strip()
-    f.close()
-    f = os.popen("fake_ifdh getToken","r") 
-    t = f.read().strip()
-    f.close()
+    p = subprocess.run('fake_ifdh getProxy', shell=True, bufsize=0, stdout=subprocess.PIPE, universal_newlines=True).stdout   
+    t = subprocess.run('fake_ifdh getToken', shell=True, bufsize=0, stdout=subprocess.PIPE, universal_newlines=True).stdout
+
```
shreyb, line 27: Is there any reason shell=True here?  Similarly, is there a reason `stdout=subprocess.PIPE` and you're not using `capture_output=True`?  Not saying I don't like it, I'm just curious as to the reason it's done this way.

Also, do we want to check the output (setting `check=True`) and raise an error (`CalledProcessError` is what the `subprocess` module will raise) if either credential call fails?

lib/condor.py
```diff
@@ -93,44 +94,52 @@ def load_submit_file(filename):
...
+
+    print("cmd_args: %s" % cmd_args)
+
+# commenting this out for now since the 'else' is not implemented
+#    if True:
+
+    cmd = "/usr/bin/condor_submit -pool %s %s %s" % (
+        COLLECTOR_HOST,
+        schedd_args,
+        " ".join(["'%s'" % x for x in cmd_args]),
```
shreyb, line 118: My understanding of the code changes you made is that if `f` is blank here, we're assuming the user provided it and it'll be passed in via `cmd_args` (like in the `condor_submit` case?

lib/condor.py
```diff
@@ -93,44 +94,52 @@ def load_submit_file(filename):
     return htcondor.Submit(res), nqueue
 
 
-def submit(f, vargs, schedd_add, cmd_args=[]):
+def submit(f, vargs, schedd_args, schedd_name, cmd_args=[]):
```
shreyb, line 97: As far as I can see, we _only_ use `schedd_name` for the setting of `_condor_CREDD_HOST` here.  From above, it seems like both in `jobsub_submit` and `condor_submit`, the schedd is already set in the `schedd_args` (always after `-name` or `-remote`).  This seems like a redundancy.

Is there a good way to pull the schedd out of the `schedd_args` so the function signature doesn't need `schedd_name`?

lib/condor.py
```diff
@@ -93,44 +94,52 @@ def load_submit_file(filename):
...
+         
+
+# This 'else' is not currently implemented
+#    else:
+#        subm, nqueue = load_submit_file(f)
+#        with schedd.transaction() as txn:
+#            cluster = subm.queue(txn, count=nqueue)
+#        print("jobid: %s@%s" % (cluster, schedd_name))
+
     return
```
shreyb, line 145: I know you didn't touch it, but do you want to make the same `os.system` ---> `subprocess.run` changes to the `submit_dag()` function below?  Or do you think we should leave it as is and change it when we templatize the dagman submission file?

shreyb (changes requested):
Looks great!  Just a couple of questions/suggestions and I think this'll be ready to go.  Thanks!

lib/condor.py
goodenou: If the user uses jobsub_submit, the file is created for them and it is passed in as f. If the user uses condor_submit (our version), then they have to pass a file in. And, as you say, it is passed in via `cmd_args`.

lib/condor.py
goodenou: Done.

lib/condor.py
shreyb: Great - thanks for clearing that up!

lib/creds.py
goodenou: Error handling done

bin/test_submit
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python3
```
shreyb, line 1: Is this file supposed to be committed to the repo?

lib/creds.py
shreyb: Thanks for looking into that!

shreyb (changes requested):
Accidentally committed file bin/test_submit?

bin/test_submit
goodenou: I guess I made this file at some point and forgot. I have removed it.

lib/condor.py
```diff
@@ -93,44 +94,59 @@ def load_submit_file(filename):
     return htcondor.Submit(res), nqueue
 
 
-def submit(f, vargs, schedd_add, cmd_args=[]):
+def submit(f, vargs, schedd_args, cmd_args=[]):
```
shreyb, line 97: So I thought about this further and discussed briefly with Kevin, and I think we're almost there with this.  Admittedly, passing the `schedd_args` and then extracting the name, as I asked you to do previously, is a little inelegant, especially since we have the `schedd_name` in both the `condor_submit` and `jobsub_submit` cases.  It'd be much better to just pass in the `schedd_name` if possible.

Given that, since the only difference between how we use the `schedd_name` comes from either using `--spool` or not in `condor_submit`, would it be possible to do the following:

1)  Change the signature here for `submit` to take `schedd_name` instead of `schedd_args`.
2)  Add to the signature a `spool` boolean argument (and default it to `False`)
3)  Take lines 129-132 from `condor_submit` and move them into this function, so we construct the `schedd_args` here.  Of course, you'd need to run the check on `spool`, not `args.spool` in that case.
4)  Change the calls to `submit` in `condor_submit` and `jobsub_submit` to pass in the `schedd_name`, and in the case of `condor_submit`, `args.spool`.

I think that'll clean this bit of code up, and help us transition to HTCondor Python bindings in the future, since we'll be passing in information, rather than strings of commands.

Please let me know if you have any questions or concerns about this.

goodenou: I made the changes. And I added in the error messages if a user tries to specify the schedd when using condor_submit

shreyb: Looks good - thanks!

shreyb (approved):

PR #37, opened 2022-05-03
Issue 31

shreyb:
Switch jobsub_lite to use tokens to authenticate to RCDS (or any other future webservice) for publishing tarballs.

### DIFF
--- a/lib/tarfiles.py
+++ b/lib/tarfiles.py
@@ -96,7 +96,7 @@ def do_tarballs(args):
 
                 cid = "".join((args.group, "%2F", digest))
                 if args.verbose:
-	                _cid = re.sub("\%2F", "/", cid)
+                    _cid = re.sub("\%2F", "/", cid)
                     print(f"Using RCDS to publish tarball\ncid: {_cid}")
 
                 publisher = TarfilePublisherHandler(cid, proxy, token)
--- a/lib/tarfiles.py
+++ b/lib/tarfiles.py
@@ -18,9 +18,11 @@ from functools import wraps
 import hashlib
 import os
 import os.path
-import requests
+import re
 import sys
 import time
+
+import requests
 
 
 def tar_up(directory, excludes):
@@ -72,7 +74,7 @@ def do_tarballs(args):
 
     NUM_RETRIES_UPLOAD = 20
     RETRY_INTERVAL_SEC = 30
-    
+
     res = []
     clean_up = []
     for tfn in args.tar_file_name:
@@ -89,25 +91,23 @@ def do_tarballs(args):
                 digest, tf = slurp_file(tfn[8:])
                 proxy, token = get_creds()
 
-                if token:
-                    with open(token, 'r') as f:
-                        token_string = f.read()
-                        token_string = token_string.strip() # Drop \n at end of token_string
-
                 if not args.group:
                     raise ValueError("No --group specified!")
+
                 cid = "".join((args.group, "%2F", digest))
-                location = pubapi_exists(cid, tokenstr=token_string) 
+
+                publisher = TarfilePublisherHandler(cid, proxy, token)
+                location = publisher.cid_exists()
                 if location is None:
-                    pubapi_publish(cid, tf, tokenstr=token_string) 
+                    publisher.publish(tf)
                     for i in range(NUM_RETRIES_UPLOAD):
                         time.sleep(RETRY_INTERVAL_SEC)
-                        location = pubapi_exists(cid, tokenstr=token_string) 
+                        location = publisher.cid_exists()
                         if location is not None:
                             break
                 else:
                     # tag it so it stays around
-                    pubapi_update(cid, tokenstr=token_string) 
+                    publisher.update_cid()
 
             elif args.use_dropbox == "pnfs":
                 location = dcache_persistent_path(args.group, tfn[8:])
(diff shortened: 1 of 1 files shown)


lib/tarfiles.py
```diff
@@ -70,6 +73,9 @@ def do_tarballs(args):
     we convert the argument to the next type as we go...
     """
 
+    NUM_RETRIES_UPLOAD = 20
```
retzkek, line 76: I'd like to see some of these configurable. Env vars maybe?

lib/tarfiles.py
```diff
@@ -128,50 +135,101 @@ def do_tarballs(args):
...
+    """Handler to publish tarballs via HTTP to RCDS (or future dropbox server)
+
+    Args:
+        object (_type_): _description_
+        cid (str): unique group/hash combination that RCDS uses to locate tarballs
+        proxy (str): Location of X509 Proxy file to authenticate to RCDS
+        token (str): Location of JWT/Sci-token to authenticate to RCDS
     """
+
     dropbox_server = "rcds.fnal.gov"
```
retzkek, line 148: Another env var?

lib/tarfiles.py
```diff
@@ -128,50 +135,101 @@ def do_tarballs(args):
...
+    def update_cid(self):
+        """Make PubAPI update call to check if we already have this tarfile
+
+        Returns:
+            requests.Response: Response from PubAPI call indicating if tarball
+            represented by self.cid is present
+        """
+        url = self.pubapi_base_url_formatter.format(endpoint="update")
+        if self.token:
+            return requests.get(url, headers=self.request_headers, verify=False)
```
retzkek, line 191: What would it take to be able to verify the server cert?

lib/tarfiles.py
retzkek: Or you can make a case why they don't need to be configurable.

shreyb: Done.  These are now configurable via env vars (I'll need to document these when I write the jobsub_lite documentation.

lib/tarfiles.py
shreyb: Done.

lib/tarfiles.py
shreyb: The problem with that is that the RCDS hostcerts don't seem to have `rcds.fnal.gov` as SANs, so the verify bit fails when we try to connect to `rcds.fnal.gov`.  I'll talk to Dave and Hyunwoo about that to see if they're OK with switching those certs out.

lib/tarfiles.py
```diff
@@ -73,8 +73,8 @@ def do_tarballs(args):
     we convert the argument to the next type as we go...
     """
 
-    NUM_RETRIES_UPLOAD = 20
-    RETRY_INTERVAL_SEC = 30
+    NUM_RETRIES_UPLOAD = os.getenv("JOBSUB_NUM_RETRIES_UPLOAD", 20)
+    RETRY_INTERVAL_SEC = os.getenv("JOBSUB_RETRY_INTERVAL_SEC", 30)
```
retzkek, line 77: These need to be converted to ints, and bad input handled (that is the downside of making things configurable...).

Also, `JOBSUB_RETRY_INTERVAL_SEC` is a bit too generic for my taste, I'd suggest starting both with `JOBSUB_UPLOAD_`.

retzkek (changes requested):

lib/tarfiles.py
```diff
@@ -82,29 +88,30 @@ def do_tarballs(args):
...
+                cid = f"{args.group}/{digest}"
+                if args.verbose:
+                    print(f"Using RCDS to publish tarball\ncid: {cid}")
+
+                publisher = TarfilePublisherHandler(cid, proxy, token)
+                location = publisher.cid_exists()
+                if location is None:
+                    publisher.publish(tf)
+                    for i in range(NUM_RETRIES_UPLOAD):
+                        time.sleep(RETRY_INTERVAL_SEC)
```
retzkek, line 108: With the default retry values Jobsub could be sitting there for up to 10 minutes waiting for a file to become available, with no feedback to the user. Can we add a message?

Also, what happens if we hit the end of the retries and the file is still not available?

lib/tarfiles.py
```diff
@@ -82,29 +88,30 @@ def do_tarballs(args):
...
+
+                # cid looks something like dune/bf6a15b4238b72f82...(long hash)
+                cid = f"{args.group}/{digest}"
+                if args.verbose:
+                    print(f"Using RCDS to publish tarball\ncid: {cid}")
+
+                publisher = TarfilePublisherHandler(cid, proxy, token)
+                location = publisher.cid_exists()
+                if location is None:
+                    publisher.publish(tf)
```
retzkek, line 106: What happens if `publish()` fails? There should be some retry logic, either here or in the `TarfilePublisherHandler` methods. Maybe the latter is preferable? There's already the `pubapi_operation` wrapper so could add retry logic there. Need to be careful about nested retries though.

retzkek (changes requested):

lib/tarfiles.py
shreyb: We don't want to do that.  Note to self:  Make sure we don't wait for cid_exists to return True before moving on if we JUST published the tarball.

lib/tarfiles.py
shreyb: Talked to Dave and Kevin, and we all agreed that mimicking the old jobsub behavior is best - that is choosing between rcds01 and rcds02 randomly, then if it fails, using the other one.  We don't expect RCDS to expand, and if it does, we have the env var already we can use to override.

lib/tarfiles.py
shreyb: Talked to Kevin - we actually will keep the logic here for now, and open a new issue for the following change:

Need to have the wrapper script do tarball discovery using the token, much like the current wrapper does.  The current way, submission will have to wait for a tarball to publish to complete.

lib/tarfiles.py
shreyb: Talked to Kevin offline, and he suggested putting retry logic into the wrapper in `TarfilePublisherHandler`.

lib/tarfiles.py
shreyb: Done.

lib/tarfiles.py
shreyb: Will be resolved in #38.

lib/tarfiles.py
shreyb: Done.

lib/tarfiles.py
shreyb: Done.  Currently, there are nested retries, but once we do #38 , that won't be an issue.

retzkek: Still need to handle if all retries of `publish()` fail, and stop the submission.

lib/tarfiles.py
```diff
@@ -128,50 +145,132 @@ def do_tarballs(args):
...
+
+        class SafeDict(dict):
+            """Use this object to allow us to not need all keys of dict when
+               running str.format_map method to do string interpolation.
+               Taken from https://stackoverflow.com/a/17215533"""
+            def __missing__(self, key):
+                return f"{{{key}}}" # "{<key>}"
+
+        def wrapper(self, *args, **kwargs):
+            _dropbox_server_selector = self.__select_dropbox_server()
```
retzkek, line 188: If I'm working through this right, we'll be trying a new server with each retry, rather than retrying a single server more than once? If that is what's desired, then is there any need for `NUM_RETRIES`? And for that matter any need for a delay between retries. IMO we should try a server more than once before moving on.

I'd also like to see a message (debug maybe?) saying what dropbox server is being tried.

retzkek (changes requested):

lib/tarfiles.py
shreyb: You're right - that's exactly how I wrote this:  So that we try a new server with each retry.  Is your thought that we should retry a single server multiple times, and if we exhaust `NUM_RETRIES`, then switch to a new server?

retzkek: That is my preference, yes, but it's worth some discussion. It would be simpler to just run through the server list, in which case you could get rid of the retry limit (now implicit in the number of servers) and maybe the interval (since you're hitting different servers, although it may help in case of transient local network issues, for example). 

I suppose in the case of only one server one could just list it multiple times...

shreyb: I'm not too married to my approach here - the way I wrote it guarantees that we hit a different server within 2 retries, but I think it's just as valid to retry a single server, then if it fails, pop it off the list and try the other for the retry limit.  I'd rather keep the retry limit, because of exactly what you said in your last point.  If there's a transient local network issue, and we have a cluster of, say, 2 servers, we'd only retry once, which may not be enough.  

Doing it like you first suggested, where we try one server for `NUM_RETRIES + 1` tries, then switch to the next one, and so on, takes longer before it fails, but it allows for some time in case of a transient issue, while ensuring that we tried each server multiple times.  So if you're OK with it, I'll keep the retry limit and interval in there, but switch the logic so we try the servers the way you suggested.

lib/tarfiles.py
shreyb: Done in the TarfilePublisherHandler wrapper.  I applied it to all the RCDS calls.

shreyb:
@retzkek - do you have any more comments on this PR?

retzkek (approved):

PR #41, opened 2022-06-01
Add --role flag support to jobsub_q

### LINKED ISSUE #22: jobsub_q doesn't support the --role flag
Make sure that something like jobsub_q -G dune --role=production <job> works

shreyb:
This PR allows us to pass the --role flag to jobsub_q, which in turn makes sure the correct bearer token is present on a non-Analysis user's machine.

Closes #22

### DIFF
--- a/bin/fake_ifdh
+++ b/bin/fake_ifdh
@@ -26,7 +26,8 @@ import os
 import time
 import argparse
 
-vault_host = "fermicloud543.fnal.gov"
+VAULT_HOST = "fermicloud543.fnal.gov"
+DEFAULT_ROLE = "Analysis"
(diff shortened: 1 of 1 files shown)


retzkek (approved):
I'd prefer to see formatting changes in a separate PR, so there's less noise to sift through here. Otherwise LGTM

PS maybe we should be using Go for all these little tools, there's a lot of shell hacks we could avoid...

shreyb:
@goodenou mentioned that I left a weak point - if `getProxy` or `getToken` was given no role, that could lead to an error condition.  I added a default role in ebd89e1e91dc.

PR #42, opened 2022-06-02
Marked /etc/condor files as %config(noreplace)

shreyb:
When we installed beta9, I realized that we accidentally wiped out the /etc/condor files, and I had to restore them from saved backups.  This should fix that.


retzkek (approved):
LGTM

PR #43, opened 2022-06-02
add InDownTime classad filter to schedd selection process

goodenou:
Looks for InDownTime classad for a schedd, and if 'True' will not choose that schedd


shreyb (approved):
Looks good.

PR #48, opened 2022-06-10
reduce use of pkg_find

marcmengel:
Cleaning out references to pkg_find everywhere except (not yet used) poms_wrapper.py.
We get everything we need for normal operation from RPM dependencies now, so finding
things in ups/Spack isn't needed.

### DIFF
--- a/bin/condor_submit
+++ b/bin/condor_submit
@@ -39,7 +39,6 @@ if PREFIX == '/usr':
 # this means we don't need fancy ups dependencies..
 #
 sys.path.append(os.path.join(PREFIX,"lib"))
-import packages
 
 #
 # import our local parts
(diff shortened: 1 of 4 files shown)


shreyb:
We're getting pushback from experiments on abandoning UPS/Spack.  Let's hold off on this for now.

shreyb:
Does this PR even apply anymore?  We should update this now that we know that we are indeed using RPMs to deploy.  Do you think it's best to start afresh or work off this PR?

marcmengel:
We could just drop this pull request, and do a new one.  At this point, it looks like it would just be removing references to module "package" in jobsub_submit.   We could then just remove lib/package.py and lib/poms_wrap.py altogether...

marcmengel:
So cancelling this one, I'll do a fresh one off current master.

Issue #10, opened 2021-12-22
Replacement for jobsub_fetchlog


shreyb:
We need in the new jobsub_lite a replacement for jobsub_fetchlog.  @retzkek is already working on this, so this is simply a placeholder issue to document the progress.

PR #53, opened 2022-06-15
Rewrite jobsub_* binaries in Go

retzkek:
Here we go...

# Summary 

This replaces the jobsub_* shell scripts and utility Python scripts with Go binaries, and moves `bin/jobsub_submit` to `lib/jobsub_submit.py`, with a `jobsub_submit` Go binary that sets up a clean Python environment to run the script in. Someday maybe we'll migrate jobsub_submit to Go, but we probably can't justify that amount of re-writing at this time. The other commands/wrappers are much simpler.

# Justification

### Experience

While Python is decidedly the _lingua franca_ in SCD, there are a number of services (e.g. Lens, FERRY) in Go and an established base of practitioners. It's also very simple to learn and onboard new developers.

The Go `os` and `os/exec` standard library packages present a very simple to use interface to common OS utilities, file operations, and subprocess spawning, so it is ergonomic to work with as a "shell script replacement" language.

Go libraries are available for easily directly working with tokens, etc, without relying on or shipping other utilities.

### Zero run-time dependencies

Maintaining Python clients (see "old" jobsub) and ensuring they work with whatever Python environment a user has configured has been an endless source of problems. Bringing in a new dependency is a major production. Go avoids that problem by being statically compiled by default, allowing the developer to fully control what dependencies are used.

### Sane argument parsing

This is not maintainable:

```bash
case "x$args" in
*--group*)
    group=`echo "$args" | sed -e 's/.*--group[=( )+]//' -e 's/ .*//'`
    args=`echo "$args*" | sed -e "s/--group[=( )+]$group//"`
    ;;
*-G*)
    group=`echo "$args" | sed -e 's/.*-G[( )+]//' -e 's/ .*//'`
    args=`echo "$args" | sed -e "s/-G[( )+]$group//"`
    ;;
esac
```

https://github.com/marcmengel/jobsub_lite/blob/master/bin/jobsub_cmd#L34=

When we're at the point of parsing arguments, we should be using a robust library to do so. See above for why Python is a bad choice for this.

### Error handling

Shell scripts are notorious for indecipherable and untraceable errors. When a user encounters an error they (and user support) should be able to identify *what* failed and *why* from the presented error messages, without needing to trace through the script to figure out what subprocess output the error. Go encourages handling errors immediately, and at every step that could go wrong, so while we can and should pass subprocess output back to the user, there will also be an explanatory message tying that to the command the user actually ran. See #4 

# Details

* Common Go functionality is put into packages under `pkg`, to be imported like `import "github.com/hepcloud/jobsub/pkg/[package]"` (note the assumed final placement for this repo - internal imports work fine regardless, since we define this in `go.mod`)

* Go binaries are created as directories under `cmd`. All binaries can be built and installed (into `$GOBIN` or `$GOPATH/bin`) with `go install ./cmd/...` (from the package root). There's a common `jobsub` command with subcommands, as well as individual `jobsub_*` binaries. Individual binaries can be built in-place like `cd cmd/jobsub; go build`.

* [`urfave/cli`](https://cli.urfave.org/v2/) was selected for argument handling, since it is flexible and fairly "light" on dependencies.

* The standard library `log` library is used for messages the user should see. This ensures pasted outputs will include timestamps. Leveled and/or structured logging is overkill and counterproductive for a user-focused command-line application (see [Dave Cheney's thoughts](https://dave.cheney.net/2015/11/05/lets-talk-about-logging
)). Instead [`clok/kemba`](https://pkg.go.dev/github.com/clok/kemba) is used to provide tagged debug logging, which can be enabled with the `DEBUG` env var, e.g. `DEBUG='*'` to print all debug logs.

# TODO

* [x] rewrite condor command wrapper
* [X] rewrite fake_ifdh to Go package
* [X] add `jobsub_fetchlog` #10 
* [ ] binary build script
* [ ] remove `bin/jobsub_*` and `bin/condor_*` (except `condor_submit`) scripts
* [ ] update RPM specfile
* [ ] test that commands work identically as "old" jobsub


retzkek:
Deferring to #55

PR #55, opened 2022-06-17
python wrappers instead of bash

marcmengel:
This is a competing request ;-)  to make the wrappers be python instead of Go. 

We are using `#!/usr/bin/python -I` to avoid environment (PYTHONHOME/PYTHONPATH) issues.


bin/jobsub_cmd
```diff
@@ -17,37 +17,65 @@
...
+            schedd = i[pos+1:]
+            i = i[0:pos]
+        execargs.append(i)
+
+    if schedd:
+        execargs.insert(0, schedd)
+        execargs.insert(0, "-name")
+
+    # also make sure we have suitable credentials...
+    os.environ['X509_USER_PROXY'] = backquote('fake_ifdh getProxy')
```
retzkek, line 69: So we're spawning a sub-process to call a python script? I assume this is to keep the python script as close to the shell as possible but I'd really rather see it done "right" now (which in my mind is making the token and proxy functions into a library) than hoping/planning to do that later.

bin/jobsub_q
```diff
@@ -1,8 +1,8 @@
-#!/bin/sh
+#!/usr/bin/python3 -I
```
shreyb, line 1: This file looks to be almost identical to jobsub_cmd (with the exception of the -global flag and calling condor_q directly).  Is there any reason jobsub_q can't just import jobsub_cmd, add the -global flag, and run that?

bin/jobsub_cmd
marcmengel: Okay, so now it imports fake_ifdh and makes the calls.

bin/jobsub_q
marcmengel: I've merged the two in the latest push; we do the -global if our command  name is jobsub_q now.

bin/jobsub_cmd
```diff
@@ -17,37 +17,79 @@
...
+
+    arglist,passthru = parser.parse_known_args()
+
+    # make list of arguments to pass to condor command:
+    # - the passthru arguments from above, except if we have
+    #   any 234@schedd style arguments, pick out the schedd and
+    #   keep the 234, and pass --name schedd as well
+    execargs = []
+    schedd = None
+    for i in passthru:
```
shreyb, line 64: What do you think about making this some sort of regex with capturing groups rather than finding "@" and splitting?  I'm worried about some rogue command where someone throws an "@" in there and it'll break this.  

That's just my personal preference, since it makes clear to future us exactly what's being parsed here, and what we're expecting.

marcmengel: That might be prettier.  Then we can also make sure it's like [\d\.]+@[\w\.]+ or something (i.e. its digits and dots then a hostname...

marcmengel: Done.

lib/fake_ifdh.py
```diff
@@ -0,0 +1 @@
+../bin/fake_ifdh
```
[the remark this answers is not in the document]
marcmengel: We're still os.system()ing it in tarfiles.py to copy the tarfiles; if we got rid of that we could get it out of bin..

bin/jobsub_cmd
```diff
@@ -17,37 +17,79 @@
...
+            i = i[0:pos]
+        execargs.append(i)
+
+    if schedd:
+        execargs.insert(0, schedd)
+        execargs.insert(0, "-name")
+
+    # also make sure we have suitable credentials...
+    role = fake_ifdh.getRole(arglist.role)
+    os.environ['X509_USER_PROXY'] = fake_ifdh.getProxy(role)
```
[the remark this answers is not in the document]
marcmengel: So I have a patch for that, but getToken() is still not really that great, as htgettoken itself exits with zero
exit status if you ask for a nonexistent group, etc.   I think I have to repeat the test for valid end date
after I call htgettoken to see if it really worked...

lib/fake_ifdh.py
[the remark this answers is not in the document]
marcmengel: Okay, cleaned that up; moved it to lib/fake_ifdh.py  Of course now we need unit tests ... ;-)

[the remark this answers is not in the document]
marcmengel: and now we have tests..

bin/jobsub_cmd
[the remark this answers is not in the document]
marcmengel: okay, now we check the token after calling htgettoken, if it fails we raise an exception.

bin/fake_ifdh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/python3 -I
```
shreyb, line 1: Sorry - I'm a bit confused.  Why do we need bin/fake_ifdh if we've got the fake_ifdh.py in lib?

lib/fake_ifdh.py
```diff
@@ -0,0 +1,155 @@
...
+    if role_override:
+        return role_override
+    elif os.environ["USER"][-3:] == "pro":
+        return "Production"
+    else:
+        return DEFAULT_ROLE 
+
+
+def checkToken(tokenfile):
+    f = os.popen("decode_token.sh -e exp %s 2>/dev/null" % tokenfile, "r")
```
shreyb, line 56: What happens if this fails? Can we check for any errors here and handle them?

lib/fake_ifdh.py
```diff
@@ -0,0 +1,155 @@
...
+    pid = os.getuid()
+    tmp = getTmp()
+    exp = getExp()
+    if exp == "samdev":
+        issuer = "fermilab"
+    else:
+        issuer = exp
+    tokenfile = "%s/bt_token_%s_%s_%s" % (tmp, issuer, role, pid)
+    os.environ["BEARER_TOKEN_FILE"] = tokenfile
+    if not checkToken(tokenfile):
```
shreyb, line 71: So I've always preferred the style of:

```
if what_I_want:
    do_something()
else:
    do_something_else()
```

What would you think about changing these lines to:

```
if checkToken(tokenfile):
    return tokenfile
else:
    all_the_other_stuff
```
? 

Not something I'm adamant on changing - I'm just curious if you agree.

lib/fake_ifdh.py
```diff
@@ -0,0 +1,155 @@
...
+        issuer = exp
+        igroup = exp
+    else:
+        issuer = "fermilab"
+        igroup = "fermilab/" + exp
+    vomsfile = "%s/x509up_%s_%s_%s" % (tmp, exp, role, pid)
+    chk_cmd = "voms-proxy-info -exists -valid 0:10 -file %s" % vomsfile
+    if 0 != os.system(chk_cmd):
+        cmd = "cigetcert -i 'Fermi National Accelerator Laboratory' -n -o %s" % certfile
+        # send htgettoken output to stderr because invokers read stdout
```
shreyb, line 105: Since this is no longer true, we should take out the comment.

lib/fake_ifdh.py
```diff
@@ -0,0 +1,155 @@
...
+        issuer = "fermilab"
+    else:
+        issuer = exp
+    tokenfile = "%s/bt_token_%s_%s_%s" % (tmp, issuer, role, pid)
+    os.environ["BEARER_TOKEN_FILE"] = tokenfile
+    if not checkToken(tokenfile):
+        cmd = "htgettoken -a %s -i %s" % (VAULT_HOST, issuer)
+        if role != DEFAULT_ROLE:
+            cmd = "%s -r %s" % (cmd, role.lower()) # Token-world wants all-lower
+        # send htgettoken output to stderr because invokers read stdout
```
shreyb, line 75: Since this is no longer true, we should take out the comment.

lib/fake_ifdh.py
```diff
@@ -0,0 +1,155 @@
...
+        res = os.system("%s >&2" % cmd)
+        if res != 0:
+            raise PermissionError("Failed attempting '%s'" % cmd)
+        if checkToken(tokenfile):
+            return(tokenfile)
+        else:
+            raise PermissionError("Failed attempting '%s'" % cmd)
```
shreyb, line 82: Love it.

shreyb (changes requested):

bin/fake_ifdh
marcmengel: So the branch will merge.   If I remove it it tells me there are conflicts to resolve.   Once we merge this, we can dink the one in bin as another pull.

lib/fake_ifdh.py
marcmengel: If this fails we have an empty expiration string and the check fails.

lib/fake_ifdh.py
marcmengel: It's all equivalent; that's just what I was thinking when I wrote it -- I do the work of getting a token if the checkToken fails.

lib/fake_ifdh.py
marcmengel: Well, I'm torn.   I'm trying to emulate ifdh, and that is its rule... We don't rely on it anymore in this code though... sigh..

marcmengel: Lets leave it.

bin/fake_ifdh
shreyb: Ah fun.  Okay - I can open an issue for that.

lib/fake_ifdh.py
shreyb: I mean in terms of any exceptions or errors running the actual `decode_token.sh` script.  If there is an exception running it, I assume we want to exit at that point (and print the error to screen), right?  If so, I'm fine leaving this as you have it.

lib/fake_ifdh.py
shreyb: Oops.  My bad - you send it to stderr, not stdout.  Sorry about the noise.

lib/fake_ifdh.py
shreyb: Yeah - I misread line 106 (and line 76) - you're sending the output to stderr, not stdout, which is what you said in the comment.  My mistake.

lib/fake_ifdh.py
marcmengel: I don't think os.popen raises anything (as long as you pass it a string); you have to check the returned handle, and the  value from the close method on the handle, to see if anything went wrong.

shreyb (approved):

PR #58, opened 2022-06-24
Forgot to add these in the  unit test branch

marcmengel:
These are the dagnabbit test files from old jobsub.


marcmengel:
RIght.  Although if anyone feels like going through and making them all different somehow, I suppose that would be good; but mostly you want to make sure they end up in the generated cmd files appropriately.

shreyb:
Great - thanks for clearing that up.  I'll go ahead and merge this.

shreyb (approved):

Issue #51, opened 2022-06-14
Implement role support for get_creds()


shreyb:
https://github.com/marcmengel/jobsub_lite/blob/8d22209b58a004bd17acc6c44e47686a765a5b60/lib/creds.py#L20

Right now, `get_creds()` will always call `fake_ifdh` with no arguments, which will allow the role-selection in `fake_ifdh` to rely on the environment.  We want to make this more intelligent, and allow for us to pass a role in.  Either check for it in `args`, or specifically allow for a `role` argument that gets passed into fake_ifdh, depending on if it's the default "Analysis" role or not.

marcmengel:
So this is half fixed overall; we call fake_ifdh methods directly now with role bits, but bin/jobsub_submit, lib/dagnabbit.py, and lib/tarfiles.py don't pass in args when they could(?) to get_creds().

PR #62, opened 2022-06-27
pass args into get_creds everywhere

marcmengel:
This should resolve issue #51, we should now be passing suitable args down into get_creds to in turn
call fake_ifdh routines to get role parameters passed down properly, instead of relying on setting variables in
the environment.


shreyb:
@marcmengel  Please resolve the merge conflict, and then we should be good to merge.

PR #63, opened 2022-06-28
Make sure jinja templates don't render if missing keys

### LINKED ISSUE #57: When rendering templates, we should be strict about checking for variables
Any time we're rendering a template, we should make sure that a variable the template expects is present in the arguments passed in.  If not, it should return an error.

This is branching off Issue #45 , and was mostly implemented by adding `undefined=jinja2.StrictUndefined` to the jinja environment.  Doing that broke a lot of the jobsub_submit unit tests, and a lot needs to be added to utils.py and test_unit.py to make the tests work again.  Most of this work is done, but thorough unit testing and general testing is needed, which is why this being split off.

shreyb:
This addresses #57.  I made the rendering of templates use `StrictUndefined` so that if the template expects a key, and the key is missing, jinja will raise an exception.  To implement this, I had to change a lot of unit tests, because they expected the old behavior.  These side-effect changes included implementing a new function to get the DN of a proxy using `voms-proxy-init` (falling back to `openssl`).

Closes #57.


marcmengel (commented):
Generally looks good.   I'm starting a discussion over in #jobsub on slack about whether we should be passing the proxy DN into the jobs anymore though.  Seems to me we should probably just drop it, and maybe add some field about what token they authenticated with instead.

PR #69, opened 2022-07-15
don't call htgettoken if we already have a good token in BEARER_TOKEN_FILE...

marcmengel:


tests/test_fake_ifdh_unit.py
```diff
@@ -68,11 +70,15 @@ def test_checkToken_success(token):
     assert res
 
 def test_getToken_good():
+    if os.environ.get('BEARER_TOKEN_FILE', None):
```
shreyb, line 73: See above.  You could use both proposed fixtures in here.

tests/test_fake_ifdh_unit.py
```diff
@@ -68,11 +70,15 @@ def test_checkToken_success(token):
...
+        del os.environ['BEARER_TOKEN_FILE']
     os.environ['GROUP'] = 'fermilab'
     token = fake_ifdh.getToken('Analysis')
     assert os.path.exists(token)
 
 def test_getToken_fail():
+    if os.environ.get('BEARER_TOKEN_FILE', None):
```
shreyb, line 80: And I suppose for this only the fixture that clears `BEARER_TOKEN_FILE`.

shreyb (changes requested):
Good - but please look at the comments and see if you agree.

tests/test_fake_ifdh_unit.py
```diff
@@ -63,16 +69,16 @@ def test_checkToken_fail():
...
     assert not res
 
-def test_checkToken_success(token):
-    res = fake_ifdh.checkToken(token)
+def test_checkToken_success(fermilab_token):
+    res = fake_ifdh.checkToken(fermilab_token)
     assert res
 
-def test_getToken_good():
+def test_getToken_good(clear_token):
```
shreyb, line 76: Shouldn't this just use your new `fermilab_token` fixture?

tests/test_fake_ifdh_unit.py
```diff
@@ -81,12 +87,12 @@ def test_getToken_fail():
     else:
         assert False
    
-def test_getProxy_good():
+def test_getProxy_good(clear_token):
```
shreyb, line 90: Shouldn't this just use your new `fermilab_token` fixture as well?

PR #72, opened 2022-07-20
Data for test jobs

marcmengel:
script to put data files in various SAM instances, test job submission with latest ifdh_version, etc.


shreyb (commented):
I don't know enough about SAM configs/metadata to judge `rebuild_gen_cfg`, but I trust that it's right.  Perhaps @retzkek  or @goodenou might know more and could provide a more useful review on that file.

Issue #46, opened 2022-06-08
Drop jobsub_transfer_data for condor_transfer_data


shreyb:
jobsub_transfer_data should be dropped from the package.  Make that condor_transfer_data, and have it do the same as jobsub_transfer_data does.  This should be rolled out at the same time as jobsub_fetchlog is implemented.

PR #73, opened 2022-07-22
jobsub_fetchlog 

### LINKED ISSUE #10: Replacement for jobsub_fetchlog

retzkek:
This adds a new `jobsub_fetchlog` binary that queries schedd to determine what local path the sandbox files will be downloaded to, then downloads the files and puts them into an archive, optionally copying them to another directory as well. Uses `htcondor` python bindings.

This adds a `Job` class in `lib/condor` that parses job IDs and handles the condor interactions.  I haven't looked at using it in other places where job ids are parsed (like the `jobsub_cmd` wrapper).

Fixes #10

### DIFF
--- a/bin/jobsub_fetchlog
+++ b/bin/jobsub_fetchlog
@@ -112,18 +112,18 @@ def main():
     else:
         # build archive
         cmd = []
-        if args.archive_format == "zip":
-            cmd = ["/usr/bin/zip", "-jq", f"{str(j)}.zip"] + [
-                os.path.join(iwd, f) for f in files
-            ]
-            # -j: junk (don't record) directory names
-            # -q: quiet
-        elif args.archive_format == "tar":
+        if args.archive_format == "tar":
             cmd = ["/usr/bin/tar", "-C", iwd, "-czf", f"{str(j)}.tgz"] + files
             # -C: move into directory so paths are relative
             # -c: create
             # -z: gzip
             # -f: filename
+        elif args.archive_format == "zip":
+            cmd = ["/usr/bin/zip", "-jq", f"{str(j)}.zip"] + [
+                os.path.join(iwd, f) for f in files
+            ]
+            # -j: junk (don't record) directory names
+            # -q: quiet
         else:
             raise Exception(f'unknown archive format "{args.archive_format}"')
         if args.verbose:
--- a/bin/jobsub_fetchlog
+++ b/bin/jobsub_fetchlog
@@ -40,12 +40,19 @@ sys.path.append(os.path.join(PREFIX, "lib"))
 # import our local parts
 #
 import creds
-from get_parser import StoreGroupinEnvironment
+from get_parser import get_base_parser
 from condor import Job
 
 
-def get_parser():
-    parser = argparse.ArgumentParser()
+def main():
+    """script mainline:
+    - parse args
+    - get credentials
+    - get job info
+    - condor_transfer_data
+    - make tarball
+    """
+    parser = get_base_parser()
     parser.add_argument("-J", "--jobid", help="job/submission ID", required=True)
     parser.add_argument(
         "--destdir",
@@ -58,32 +65,7 @@ def get_parser():
         help='format for downloaded archive: "tar" (default, compressed with gzip) or "zip"',
         default="tar",
     )
-    parser.add_argument(
-        "-G",
-        "--group",
-        help="Group/Experiment/Subgroup for priorities and accounting",
-        action=StoreGroupinEnvironment,
-        default=os.environ.get("GROUP", None),
-    )
-    parser.add_argument(
-        "--verbose",
-        action="store_true",
-        default=False,
-        help="dump internal state of program (useful for debugging)",
-    )
 
-    return parser
-
-
-def main():
-    """script mainline:
-    - parse args
-    - get credentials
-    - get job info
-    - condor_transfer_data
-    - make tarball
-    """
-    parser = get_parser()
     args = parser.parse_args()
 
     if args.verbose:
--- a/bin/jobsub_fetchlog
+++ b/bin/jobsub_fetchlog
@@ -20,7 +20,6 @@
 import argparse
 import sys
 import os
-import os.path
 import subprocess
 import shutil
 
@@ -93,37 +92,45 @@ def main():
     try:
         os.stat(iwd)
     except FileNotFoundError:
-        os.mkdir(iwd, mode=0o750)
+        os.makedirs(iwd, mode=0o750)
 
     # get the output sandbox
     j.transfer_data()
+    files = os.listdir(iwd)
 
-    owd = iwd
-    # if the user wants output in a specific directory, copy files there
     if args.destdir is not None:
+        # If the user wants output in a specific directory, copy files there,
+        # don't build an archive. Old jobsub would get an archive from the
+        # server, upack it into the dest dir, then delete the archive.
         owd = args.destdir
-        shutil.copytree(iwd, owd)  # TODO python3.9+: dirs_exist_ok=True?
-
-    # build tarball
-    files = os.listdir(owd)
-    cmd = []
-    if args.archive_format == "zip":
-        cmd = ["/usr/bin/zip", "-jq", f"{str(j)}.zip"] + [owd + "/" + f for f in files]
-        # -j: junk (don't record) directory names
-        # -q: quiet
-    elif args.archive_format == "tar":
-        cmd = ["/usr/bin/tar", "-C", owd, "-czf", f"{str(j)}.tgz"] + files
-        # -C: move into directory so paths are relative
-        # -c: create
-        # -z: gzip
-        # -f: filename
+        try:
+            os.stat(owd)
+        except FileNotFoundError:
+            os.makedirs(owd, mode=0o750)
+        for f in files:
+            shutil.copy2(os.path.join(iwd, f), owd)  # copy2 tries to preserve metadata
     else:
-        raise Exception(f'unknown archive format "{args.archive_format}"')
-    if args.verbose:
-        print(f'running "{cmd}"')
-    p = subprocess.Popen(cmd)
-    if p.wait() != 0:
-        raise Exception(f"error creating archive")
+        # build archive
+        cmd = []
+        if args.archive_format == "zip":
+            cmd = ["/usr/bin/zip", "-jq", f"{str(j)}.zip"] + [
+                os.path.join(iwd, f) for f in files
+            ]
+            # -j: junk (don't record) directory names
+            # -q: quiet
+        elif args.archive_format == "tar":
+            cmd = ["/usr/bin/tar", "-C", iwd, "-czf", f"{str(j)}.tgz"] + files
+            # -C: move into directory so paths are relative
+            # -c: create
+            # -z: gzip
+            # -f: filename
+        else:
+            raise Exception(f'unknown archive format "{args.archive_format}"')
+        if args.verbose:
+            print(f'running "{cmd}"')
+        p = subprocess.Popen(cmd)
+        if p.wait() != 0:
+            raise Exception(f"error creating archive")
 
 
 if __name__ == "__main__":
--- a/bin/jobsub_fetchlog
+++ b/bin/jobsub_fetchlog
@@ -1,0 +1,134 @@
+#!/usr/bin/python3 -I
+
+#
+# jobsub_fetchlog -- tool for downloading job output files from condor
+# COPYRIGHT 2022 FERMI NATIONAL ACCELERATOR LABORATORY
+#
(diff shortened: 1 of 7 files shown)


bin/jobsub_fetchlog
```diff
@@ -0,0 +1,148 @@
...
+    - condor_transfer_data
+    - make tarball
+    """
+    parser = get_parser()
+    args = parser.parse_args()
+
+    if args.verbose:
+        htcondor.set_subsystem("TOOL")
+        htcondor.param["TOOL_DEBUG"] = "D_FULLDEBUG"
+        htcondor.enable_debug()
```
retzkek, line 92: This can be a bit chatty, but I think we should consider setting these options for all commands when the verbose/debug flag is set, otherwise the condor commands can be black holes.

bin/jobsub_fetchlog
```diff
@@ -0,0 +1,148 @@
...
+        help='format for downloaded archive: "tar" (default, compressed with gzip) or "zip"',
+        default="tar",
+    )
+    parser.add_argument(
+        "-G",
+        "--group",
+        help="Group/Experiment/Subgroup for priorities and accounting",
+        action=StoreGroupinEnvironment,
+        default=os.environ.get("GROUP", None),
+    )
```
retzkek, line 67: This is just copied from lib/get_parser,  do we want to abstract out "common" args? And does it make sense to keep all the jobsub_submit-specific args in lib/get_parser?

bin/jobsub_fetchlog
```diff
@@ -0,0 +1,148 @@
...
+        os.mkdir(iwd, mode=0o750)
+
+    # get the output sandbox
+    j.transfer_data()
+
+    owd = iwd
+    # if the user wants output in a specific directory, copy files there
+    if args.destdir is not None:
+        owd = args.destdir
+        shutil.copytree(iwd, owd)  # TODO python3.9+: dirs_exist_ok=True?
```
retzkek, line 123: This will fail if the destination directory already exists, is that what we want? Configurable in later pythons than we are currently targetting.

lib/condor.py
```diff
@@ -188,3 +189,81 @@ def submit_dag(f, vargs, schedd_name, cmd_args=[]):
...
+        123.0@schedd.example.com
+        123.456@schedd.example.com
+
+    """
+
+    id: str
+    seq: str
+    proc: str
+    schedd: str
+    cluster: bool
```
retzkek, line 213: I think this is the first place in the codebase to add type annotations. Thoughts?

bin/jobsub_fetchlog
shreyb: I wrestled with the same question a couple months ago when doing some work on jobsub_submit, and at the time there wasn’t enough motivation to abstract out common args. I think now there is.  Something like having a common parser in get_parser, import it here, and add to it?

bin/jobsub_fetchlog
shreyb: Agreed, and I’ll open up another issue to clean up some of the too-verbose output that otherwise appears, so that there’s little noise without verbose, and lots of info with,

bin/jobsub_fetchlog
shreyb: My take is that we should probably be okay if the directory already exists. Maybe make that work for now, keep the note for when we move to python 3.9?

lib/condor.py
shreyb: Why are seq and proc str and not int?

lib/condor.py
```diff
@@ -188,3 +189,81 @@ def submit_dag(f, vargs, schedd_name, cmd_args=[]):
             print("Execution failed: ", e)
 
     return submit(subfile, vargs, schedd_name)
+
+
+class Job:
```
shreyb, line 194: Should the class be Job if we’re including clusters? Aren’t all jobs clusters, and not the reverse? What are your thoughts?

bin/jobsub_fetchlog
retzkek: I agree. Here's what I've identified as the "common" args:

* -G/--group
* --role
* --subgroup (?)
* --verbose

Anything else you can think of? I'll go ahead and add that change to this PR.

marcmengel: > This is just copied from lib/get_parser, do we want to abstract out "common" args? And does it make sense to keep all the jobsub_submit-specific args in lib/get_parser?

We could add a defaulted argument to get_parser() to have it return just the common subset, and a second to add the pass-through list that the jobsub/condor wrapper uses...  So it would be `get_parser(just_common=False, add_passthru=False)`

shreyb: Is the set of jobsub/condor args pretty constant when it's not the common set? If so, this sounds good to me. If not, I'd advocate for having the extra args for each jobsub/condor command in those files themselves.

marcmengel (approved):
I think with the discussed changes, this will be fine.

lib/condor.py
marcmengel: I think we call it "jobsub-job-id" all over the place, even when we're referring to a cluster, so calling the class Job seems okay to me.

lib/condor.py
marcmengel: I think type annotations are  good; I certainly wouldn't take them out.   Maybe we should make an issue to go back through and type-annotate all our variables?

bin/jobsub_fetchlog
retzkek: I added a `get_base_parser()` function with the general args in a group, which `get_parser()` expands on for submit. It's a bit backwards maybe, but renaming `get_parser()` would impact a number of packages and tests.

bin/jobsub_fetchlog
retzkek: Reworked doing individual file copies instead of `copytree()`, there shouldn't be any subdirs anyways.

lib/condor.py
retzkek: Changed seq and proc to ints.

retzkek:
Requested changes made and rebased onto `master` (plus some commit cleanup).

shreyb:
@marcmengel I re-requested your review because it's from 7/26, long before Kevin made a bunch of new changes.  So we should re-review this.

retzkek:
Oh let me take care of #46 too.

marcmengel (approved):
Looks very nice. 
I should update jobsub_cmd now to use get_base_parser; but that could be a separate PR.

retzkek:
I removed `jobsub_transfer_data` and updated integration test to use `jobsub_fetchlog` instead. I also removed the "output will be in..." messages from submit, which users don't need (and I don't think we really want them to know) since `jobsub_fetchlog` abstracts that away.

Updated specfile.

Integration tests running overnight.

retzkek:
I ran a simplified integration test of just two test jobs, which ran OK. Modified tests (just commenting out most of the launch "tests") not committed - we should refactor the integration tests but I'll leave that for another time.

```
=============================================================================== test session starts =================================================================================
platform linux -- Python 3.6.8, pytest-7.0.1, pluggy-1.0.0 -- /home/kretzke/jobsub_lite/venv/bin/python3
cachedir: .pytest_cache
rootdir: /home/kretzke/jobsub_lite/tests, configfile: pytest.ini
collected 5 items                                                                                                                                                                    

tests/test_submit1.py::test_launch_lookaround_samdev PASSED                                                                                                                    [ 20%]
tests/test_submit1.py::test_launch_lookaround_dune PASSED                                                                                                                      [ 40%]
tests/test_submit1.py::test_wait_for_jobs PASSED                                                                                                                               [ 60%]
tests/test_submit1.py::test_fetch_output PASSED                                                                                                                                [ 80%]
tests/test_submit1.py::test_check_job_output PASSED                                                                                                                            [100%]

=========================================================================== 5 passed in 832.55s (0:13:52) ============================================================================
```

lib/condor.py
shreyb: Looks good.

lib/condor.py
```diff
@@ -227,12 +221,93 @@ def submit_dag(
...
+            self.seq = int(m.group(1))  # seq is required
+            if m.group(2) is None:  # proc is optional
+                self.cluster = True
+                self.proc = 0
+            else:
+                self.cluster = False
+                self.proc = int(m.group(2))
+            self.schedd = m.group(3)  # schedd is required
+        except TypeError as e:
+            raise JobIdError(f'error when parsing job id "{job_id}"') from e
```
shreyb, line 272: Nice - we need some more of these custom errors in the code as we move closer to production.

lib/condor.py
```diff
@@ -227,12 +221,93 @@ def submit_dag(
...
+
+    def __str__(self) -> str:
+        if self.cluster:
+            return f"{self.seq}@{self.schedd}"
+        return f"{self.seq}.{self.proc}@{self.schedd}"
+
+    def _get_schedd(self) -> htcondor.htcondor.Schedd:
```
shreyb, line 279: Do we anticipate using these protected methods outside of the `Job` class?  If not, this and `_constraint` should probably be protected, as in `__get_schedd` and `__constraint`

lib/get_parser.py
```diff
@@ -44,9 +44,37 @@ def __call__(
         setattr(namespace, self.dest, values)
 
 
-def get_parser() -> argparse.ArgumentParser:
-    """build the argument parser and return it"""
+def get_base_parser() -> argparse.ArgumentParser:
```
shreyb, line 47: Yeah, I really don't like this as a long-term solution, but I understand why you did it. IMO, the proper way to do it is to have this base_parser, and have each executable only build-in the arguments it needs on top of the base parser, in the executable files themselves.  I'm fine with making that a separate ticket if you agree.

tests/test_condor_unit.py
```diff
@@ -145,3 +145,36 @@ def test_submit_dag_1(self, get_dag_file, needs_credentials):
             get_dag_file, TestUnit.test_vargs, TestUnit.test_schedd, cmd_args=[]
         )
         assert res
+
+
+class TestJob:
+    def test_job(self):
```
shreyb, line 151: Could you please mark all of these unit tests with the `@pytest.mark.unit` decorator?  We want to have a github action where all the unit tests are run for each PR, and that'll help us separate unit, system, and integration tests.

shreyb (changes requested):
This looks great.  I only have two real requests for changes:

1.  In `test_condor_unit.py`, to mark the new tests as unit tests (see comment)
2. Rebase to make sure we don't blow away Lisa's changes from yesterday, since they're in `get_parser.py` as well.  I'm not sure this item is necessary, but just to be safe.

The rest of my comments (of which there are a lot) are VERY nitpicky - feel free to take them or ignore them.

bin/jobsub_fetchlog
```diff
@@ -0,0 +1,137 @@
...
+            os.stat(owd)
+        except FileNotFoundError:
+            os.makedirs(owd, mode=0o750)
+        for f in files:
+            shutil.copy2(os.path.join(iwd, f), owd)  # copy2 tries to preserve metadata
+    else:
+        # build archive
+        cmd = []
+        if args.archive_format == "zip":
+            cmd = ["/usr/bin/zip", "-jq", f"{str(j)}.zip"] + [
```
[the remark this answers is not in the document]
retzkek: That's a fair comment, simple enough.

lib/condor.py
retzkek: Agreed. We should also discuss overall error handling and what to show the user - python tracebacks are informative to developers (and so good in tickets) but are not so helpful to the users. But, later...

lib/condor.py
retzkek: I decided single underscore was appropriate, since they're not intended to be used outside the class, but there's no reason they couldn't, and they have no side effects. But since you brought it up, PEP8 also says

> Generally, double leading underscores should be used only to avoid name conflicts with attributes in classes designed to be subclassed.

Which `Job` isn't.

tests/test_condor_unit.py
retzkek: OK

lib/get_parser.py
retzkek: OK. I'll push the other changes so you can review, then rebase again.

lib/condor.py
shreyb: Fair enough.

lib/get_parser.py
shreyb: Everything looks good - go ahead and rebase please.

shreyb (approved):
Looks good - with the caveat that this will be rebased on master before merging.

shreyb:
This looks good to me.

PR #74, opened 2022-07-27
first stab at type annotations

marcmengel:
Type annotations on functions/args throughout.

### DIFF
--- a/bin/condor_submit
+++ b/bin/condor_submit
@@ -50,7 +50,7 @@ COLLECTOR_HOST = htcondor.param.get("COLLECTOR_HOST", "gpcollector03.fnal.gov")
 class StoreGroupinEnvironment(argparse.Action):
     """Action to store the given group in the GROUP environment variable"""
 
-    def __call__(self, parser, namespace, values, option_string=None):
+    def __call__(self, parser:argparse.ArgumentParser, namespace:Any, values:Dict[str,str], option_string=None:Union[None,str]):
         os.environ["GROUP"] = values
         setattr(namespace, self.dest, values)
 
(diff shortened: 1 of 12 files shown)


bin/condor_submit
```diff
@@ -50,7 +50,7 @@ COLLECTOR_HOST = htcondor.param.get("COLLECTOR_HOST", "gpcollector03.fnal.gov")
 class StoreGroupinEnvironment(argparse.Action):
     """Action to store the given group in the GROUP environment variable"""
 
-    def __call__(self, parser, namespace, values, option_string=None):
+    def __call__(self, parser:argparse.ArgumentParser, namespace:Any, values:Dict[str,str], option_string=None:Union[None,str]):
```
shreyb, line 53: Any reason the `namespace` is type `Any` when we're using this with `argparse`, so the expected type is `argparse.Namespace`?

bin/jobsub_cmd
```diff
@@ -31,17 +31,12 @@ import fake_ifdh
 class StoreGroupinEnvironment(argparse.Action):
     """Action to store the given group in the GROUP environment variable"""
 
-    def __call__(self, parser, namespace, values, option_string=None):
+    def __call__(self, parser: argparse.ArgumentParser, namespace:Any, values:dict[str,str], option_string=None:Union[None,str]):
```
shreyb, line 34: Same question as `condor_submit`

shreyb (changes requested):
Looks great.  I just had a couple of questions.

bin/condor_submit
marcmengel: Mostly I didn't chase down what type it actually was... I can update that.

shreyb (approved):

PR #78, opened 2022-08-05
Added pre-commit config and ran entire code base through it

shreyb:
Added pre-commit config with black, pylint, and mypy tests, along with a few other hooks.

Lots and LOTS of changes, so reviewers, please take your time as needed.


shreyb:
Moved to draft status because of a couple of testing issues.

shreyb:
Fixed the testing issues, but I realized a couple of my unit tests fail because of fife_launch and POMS things.  Will discuss with @marcmengel  before moving this to Ready for Review.

shreyb:
Marc agreed that /grid/fermiapp-dependent code might come out in the future anyhow (we're discussing offline).  For now, I mounted /grid/fermapp to my dev node, and the package unit tests work.  So I'm going to move this to "ready".

shreyb:
Reviewers:  Please note that most of the changes are whitespace/EOF changes.

marcmengel (approved):
Looks good.  Only thing I would have done differently is just made all the units conversion tables in utils.py be Dict[str,float] with a few strategic .0's added.

shreyb:
Thanks Marc - I'll do what you said and change the units conversion tables before merging this then (and I'll rerun tests of course).

shreyb:
Made Marc's suggested changes.  @retzkek, @goodenou do either of you have any comments before I merge this in (next week, naturally)?

goodenou (approved):
I reviewed the "tests" directory. Changes all look good.

lib/fake_ifdh.py
```diff
@@ -62,7 +62,7 @@ def checkToken(tokenfile: str) -> bool:
     cmd = f"decode_token.sh -e exp {tokenfile} 2>/dev/null"
     with os.popen(cmd, "r") as f:
         exp_time = f.read()
-    return exp_time and ((int(exp_time) - time.time()) > 60)
+    return bool(exp_time) and ((int(exp_time) - time.time()) > 60)
```
retzkek, line 65: `bool(exp_time)` isn't really the right answer here, although "should" mostly work. Actually I'm not sure what the point of that guard is, all it will do is make it even harder to debug issues since it will swallow any errors from running decode_token, which we don't otherwise check for (but should). Instead we'd get a message about a problem running `htgettoken` which would be less than helpful. But I digress.

lib/tarfiles.py
```diff
@@ -83,7 +84,7 @@ def dcache_persistent_path(exp: str, filename: str) -> str:
     return res
 
 
-def do_tarballs(args: Dict[str, str]) -> None:
+def do_tarballs(args: argparse.Namespace) -> None:
```
retzkek, line 87: Should we be leaking the `argparse` or would it be better to pass along specific arguments?

lib/tarfiles.py
```diff
@@ -92,7 +93,7 @@ def do_tarballs(args: Dict[str, str]) -> None:
     """
     # pylint: disable=too-many-nested-blocks,too-many-branches
     res = []
-    clean_up = []
+    clean_up: List[Any] = []
```
retzkek, line 96: We know `clean_up` is a `List[str]` right, since `tar_up()` returns `str`?

lib/tarfiles.py
```diff
@@ -145,7 +146,7 @@ def do_tarballs(args: Dict[str, str]) -> None:
         try:
             os.unlink(tf)
         except:  # pylint: disable=bare-except
-            print(f"Notice: unable to remove generated tarfile {tf}")
+            print(f"Notice: unable to remove generated tarfile {tf}")  # type: ignore
```
retzkek, line 149: Why the `type: ignore` here?

retzkek (changes requested):

lib/fake_ifdh.py
shreyb: Okay - I think I resolved this in the next commit.  Good catch - this is a big issue as written (and as I modified it).  For example:

```python
def func():
    return "foo" and (int('') > 3)

print(func())
``` 
raises the correct `ValueError`, but the following:

```python
def func():
    return '' and (int('') > 3)

print(func())
```
gives an empty string.  So this is a bug we've been masking this whole time.  I've updated the code and the unit test.

lib/tarfiles.py
shreyb: Yeah - it looks like mypy has a few open bugs where it confuses `bytes` and `strings` in certain cases.  See 
https://github.com/python/mypy/issues/12961
and 
https://github.com/python/mypy/issues/12666

This seems to be related, as the output of pre-commit if I make this and the below change is that I'm trying to use the product of an iteration (`str`) as a `byte` object (Line 145, 149).  So that's why I've set the type of `clean_up` to be `List[Any]` and ignored line 149 (which produces an identical error to the second issue I've linked above).

lib/tarfiles.py
shreyb: See above comment about possible mypy open bugs.

lib/tarfiles.py
shreyb: The reason we're passing in `args` here is because so much of the code-base, like previous jobsub, is written to operate on the entire Namespace and pick what it needs.  We can certainly change that, but it'd require a significant rewrite (we're talking at least most of the credentials stuff, condor stuff, etc.).  I'd be happy to revisit that, and perhaps we should, but what are your thoughts on opening a new issue for that?

lib/tarfiles.py
shreyb: Never mind.  Fixed.  It was type inference.  We used `tf` as a `bytes` earlier, and then when reassigning, used it as a `str` in a loop later (145), and mypy wasn't having that.

Reference:  https://mypy.readthedocs.io/en/stable/type_inference_and_annotations.html

So I made both this change and the one suggested below in [255129e](https://github.com/marcmengel/jobsub_lite/pull/78/commits/255129ebada73a564b2cabbad9733d5ff46bdfc8).

lib/tarfiles.py
retzkek: Yeah that's fair. Don't bother with an issue, maybe someone will have a reason refactor that in the future, or maybe not.

lib/fake_ifdh.py
retzkek: Yes, exactly. Worst case for most errors here we'd just end up getting a new token anyways, which I suppose was the initial thinking, but it's those couple other error conditions where everything goes haywire that we need to catch.

Your fixes look good, I like the additional error text.

lib/tarfiles.py
retzkek: Good catch!

retzkek (approved):
LGTM

shreyb:
Thanks to all three of you for reviewing this monumental PR.  Please go ahead and set up pre-commit (https://pre-commit.com/), and if you need help with it, let me know.

Issue #75, opened 2022-08-02
Implement jobsub_lite schedd classad


shreyb:
Nick Peregonow created a classad for jobsub_lite schedds.  We need to implement this in jobsub_lite.
Classad is "SCHEDD.IsJobsubLite = True" for jobsub_lite schedd picker.

shreyb:
Started work for this here:
https://github.com/shreyb/jobsub_lite/tree/issue_75

Need to run unit tests, but it looks like the classad is NOT implemented:

```
$ condor_status -schedd -af Name IsJobsubLite
gpce03.fnal.gov undefined
gpce04.fnal.gov undefined
jobsub01.fnal.gov undefined
jobsub02.fnal.gov undefined
jobsub03.fnal.gov undefined
jobsubdevgpvm01.fnal.gov undefined
```

Will need to keep this on hold until Nick comes back, because we should test it before opening a PR.

PR #79, opened 2022-08-12
Use IsJobsubLite classad to select jobsub_lite schedds

shreyb:
This change will make jobsub_lite only select the schedds in the pool where IsJobsubLite is set to "true".  It works in conjunction with the `--devserver` flag, so if that flag is specified, only jobsub_lite devservers are chosen, and if it is not, then only non-dev jobsub_lite schedds are chosen.

### DIFF
--- a/lib/condor.py
+++ b/lib/condor.py
@@ -45,7 +45,8 @@ def get_schedd(vargs: Dict[str, str]) -> classad.ClassAd:
     # need to directQuery for the full classads to check for
     # SupportedVOList...
 
-    print("classads: ", schedd_classads)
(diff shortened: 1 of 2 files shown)


shreyb:
Closes #75

goodenou (approved):
Looks good to me. What is the reason behind wanting to have schedds that don't run jobsub_lite? Is this for the testing phase?

shreyb:
Sorry if that was unclear, @goodenou - with this PR, we only pick jobsub_lite schedds, regardless of whether we use `--devserver` or not.

PR #80, opened 2022-08-15
make more things debug messages...

marcmengel:
…  and run htgettoken, voms-proxy-init, etc, output to /dev/null unless debug.

This is an attempt to close out #4.


shreyb (approved):
This looks good.  I think #4 is a little too vague anyhow, so we should start filing specific issues for what output we actually want to suppress, and which should come through debug.

shreyb:
Sorry Marc - I knew this might happen with the piling up PRs.  Would you please mind resolving these conflicts at your convenience?

PR #85, opened 2022-08-21
Dropbox in dags

marcmengel:
Test cases for submitting dagnabbit dags, including ones with --tar-file-name and -f drobpox:something.
Along with assorted code fixes to make the tests actually work.
This turned up several issues, the most festive was that since we wait for the transfers, the later DAG stages made a new submit directory (because the date-stamp changed) and so the files we needed for submission got spread amongst multiple directories...


lib/tarfiles.py
```diff
@@ -92,8 +92,25 @@ def do_tarballs(args: argparse.Namespace) -> None:
...
     """
     # pylint: disable=too-many-nested-blocks,too-many-branches
-    res = []
     clean_up: List[str] = []
+    res: List[str] = []
+    path: Optional[str] = None
+    for fn in args.input_file:
+        if fn.startswith("dropbox:"):
+            tarfile = tar_up(
+                os.path.dirname(fn[8:]), "/dev/null", os.path.basename(fn[8:])
```
shreyb, line 101: I've never been a fan of substringing like this (`fn[8:]`) in these cases, just because it makes the reader have to do extra mental work to figure out what's at position 8, and therefore what the resultant string would be.  What would you think about using replace (e.g. fn.replace("dropbox:", "") or a regex?  If you agree, I'll put in an issue to do the same elsewhere in the code (for example, on line 117 for `tardir:`).

shreyb (commented):
Looks good - I just had one stylistic comment.

marcmengel:
Okay, but while I was looking at that, I realized I wasn't handling the pnfs dropbox: case right...

shreyb (approved):
Thanks Marc - looks good.

Issue #84, opened 2022-08-18
Implement clearing out of troublesome environment variables


shreyb:
There are a number of known environment variables that cause issues when running grid jobs in singularity containers.  We should, by default, clear those, and provide a flag to keep those variables in place.

This list of variables should be _well-documented_.

marcmengel:
Pull request: #87

Added writeup on wrapper script to local wiki: https://github.com/marcmengel/jobsub_lite/wiki/Jobsub-wrapper-script
we should get this moved to the fife_wiki. https://fifewiki.fnal.gov/wiki/

PR #87, opened 2022-08-22
Clearing problem environment variables

marcmengel:
This is for issue #84


shreyb (approved):
Do we expect this list to get very long?  If so, we might want to define an array and then loop over that just so it looks cleaner.  If it's, say, under 5 vars, this is perfectly fine IMO.

PR #93, opened 2022-08-24
remove EXPERIMENT variable in favor of GROUP

goodenou:
We agreed to use the variable GROUP exclusively in jobsub code to define the VO. I removed references to the other variables with the same meaning.

### DIFF
--- a/lib/fake_ifdh.py
+++ b/lib/fake_ifdh.py
@@ -39,9 +39,8 @@ def getTmp() -> str:
 
 def getExp() -> Union[str, None]:
     """return current experiment name"""
-    for ev in ["GROUP", "EXPERIMENT", "SAM_EXPERIMENT"]:
-        if os.environ.get(ev, None):
-            return os.environ.get(ev)
+    if os.environ.get("GROUP", None):
+        return os.environ.get("GROUP")
     # otherwise guess primary group...
     exp = None
     with os.popen("id -gn", "r") as f:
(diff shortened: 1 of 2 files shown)


tests/test_submit_wait_int.py
```diff
@@ -46,7 +44,6 @@ def noexp(job_envs):
 def samdev(job_envs):
     """fixture to run launches for samdev/fermilab"""
     os.environ["GROUP"] = "fermilab"
-    os.environ["EXPERIMENT"] = "samdev"
```
shreyb, line 49: Doesn't IFDH need these to work?  Or do we not need them for the tests to pass, but since the job wrappers add `EXPERIMENT`, we're OK?

@marcmengel - do you have any input on deleting these from tests?

marcmengel: We do need the EXPERIMENT set in the tests so any SAM stuff can work in the jobs.

shreyb (changes requested):
Okay - I was afraid of that.  Please keep the EXPERIMENT hardcodings in the SAM tests and fife_launch test.

tests/test_submit_wait_int.py
goodenou: OK. I will put "EXPERIMENT" back into the tests.

shreyb (approved):
Looks good - thanks!

PR #97, opened 2022-08-30
default role from token

marcmengel:
So, much like old jobsub takes default role from a proxy if present, this takes the default role from
a token in BEARER_TOKEN_FILE.

### DIFF
--- a/bin/decode_token.sh
+++ b/bin/decode_token.sh
@@ -9,11 +9,16 @@
 decode_token() {
     #
     # token is 3 base64 segments separated by dots
-    # we want the middle one.
-    # the ending of the base64 segment is sometimes
-    # chopped, so ignore errors about that
+    # we want the middle one
     #
-    sed -e 's/.*\.\(.*\)\..*/\1==/' "$1" 2>/dev/null | base64 -d  2>/dev/null
+    sed -e 's/.*\.\(.*\)\..*/\1==/' "$1" | base64 -d 2>/dev/null
+}
+
+pretty_print() {
+   # sed regexps:
+   #  - break line and indent after (some) commas,
+   #  - put braces on their own line
+   sed -e 's/:[^:]*[]"0-9],/&\n  /g' -e 's/[{}]/\n&\n  /g'
 }
 
 get_field() {
@@ -25,7 +30,7 @@ get_field() {
     # BUGS
     #    Borked for fields that contain commas.
     #
-    sed -e "s/.*\"$1\"://" -e s'/,.*//'
+    sed -e "s/.*\"$1\"://" -e s'/,"[^"]*":.*//'
 }
 
 usage() {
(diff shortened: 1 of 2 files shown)


lib/fake_ifdh.py
```diff
@@ -50,10 +50,22 @@ def getExp() -> Union[str, None]:
...
-        return "Production"
+
+    # if there's a role in the wlcg.groups of the token, pick that
+    if os.environ.get("BEARER_TOKEN_FILE", False):
+        with os.popen("decode_token.sh $BEARER_TOKEN_FILE", "r") as f:
+            token_s = f.read()
+            token = json.loads(token_s)
+            groups: List[str] = token["wlcg.groups"]
+            if len(groups) == 2:
+                pos = groups[1].find("/", 2)
```
shreyb, line 64: As we discussed on slack, it might be more robust to make this a regex capturing group.  We know that the groups will look something like this:

```
    "/dune",
    "/dune/production"
```
So rather than assume that the production (or non-default role) will always come second, check each key against a regex; roughly:

```python
experiment = getExp()
role_regex = re.compile(f"\/{experiment}\/(\w+)"))
check_for_subgroup_match()
return subgroup_match
```

lib/fake_ifdh.py
```diff
@@ -50,10 +50,22 @@ def getExp() -> Union[str, None]:
...
         return role_override
-    if os.environ["USER"][-3:] == "pro":
-        return "Production"
+
+    # if there's a role in the wlcg.groups of the token, pick that
+    if os.environ.get("BEARER_TOKEN_FILE", False):
+        with os.popen("decode_token.sh $BEARER_TOKEN_FILE", "r") as f:
+            token_s = f.read()
+            token = json.loads(token_s)
+            groups: List[str] = token["wlcg.groups"]
```
shreyb, line 62: Can we either have this be a `token.get` call that will return `DEFAULT_ROLE` if it finds no `wlcg.groups` key, or a `try-except` block that does the same if it catches `KeyError`?  Just in case our token is malformed.

bin/decode_token.sh
```diff
@@ -9,11 +9,16 @@
...
     #
-    sed -e 's/.*\.\(.*\)\..*/\1==/' "$1" 2>/dev/null | base64 -d  2>/dev/null
+    sed -e 's/.*\.\(.*\)\..*/\1==/' "$1" | base64 -d 2>/dev/null
+}
+
+pretty_print() {
+   # sed regexps:
+   #  - break line and indent after (some) commas,
+   #  - put braces on their own line
+   sed -e 's/:[^:]*[]"0-9],/&\n  /g' -e 's/[{}]/\n&\n  /g'
```
shreyb, line 21: Pardon my lack of `sed` knowledge, but in the first substitution, is there an extra `]` in the search part?  Or is it missing an opening `[` before `0-9`?  Or am I just wrong?

Also, I assume that you're adding the pretty_print() function because you don't want us to rely on the installation of `jq`, right?

shreyb (changes requested):

Issue #98, opened 2022-08-30
If we know the schedd, we should set _condor_CREDD_HOST to it in the environment


shreyb:
Currently, `jobsub_submit` automatically sets the environment variable `_condor_CREDD_HOST` based on the schedd it picks for submission.  We need to ensure that for lightly-wrapped condor commands like jobsub_transfer_data, jobsub_rm, etc., where we know the schedd _a priori_ because the user has to specify it, that we set `_condor_CREDD_HOST` to the schedd.  Otherwise, we run into OIDC issues with the token where a user has to reauthenticate because `condor_vault_storer` (run by all condor commands) thinks there's no credentials stored.  I believe the line referenced here is where we would have to set the environment variable:

https://github.com/marcmengel/jobsub_lite/blob/dbd77b3bb04a0c98b00aa0f32812cea478582a1e/bin/jobsub_cmd#L76

Reproduce issue:

On a managed tokens node:
```
[dunepro@hostname ~]$ export _condor_COLLECTOR_HOST=<collector_host>
[dunepro@hostname ~]$ export HTGETTOKENOPTS="--credkey=<credkey>"
[dunepro@hostname ~]$ condor_vault_storer -v dune_production

Checking if -s dune_production credentials exist
Account: <current> (dunepro)
CredType: oauth

Operation failed.
    Make sure your ALLOW_WRITE setting includes this host.
Removing /tmp/vt_u50762-dune_production because there are no dune_production credentials stored
Attempting to get tokens for dune_production
Initializing kerberos client for host@<vault_server>
Negotiating kerberos with <vault_server>
  at path auth/kerberos-dune_production
Attempting to get bearer token from <vault_server>
  at path <path>
Read token from <URL> failed: HTTPError: HTTP Error 403: Forbidden: 1 error occurred:
        * permission denied


No ssh-agent keys found
htgettoken: Failure getting token from <vault server>
Authentication needed for dune_production
Initializing kerberos client for <vault server>
Negotiating kerberos with <vault server>
  at path auth/kerberos-dune_production
Attempting to get bearer token from <vault server>
  at path <path>
Read token from <path> failed: HTTPError: HTTP Error 403: Forbidden: 1 error occurred:
        * permission denied


No ssh-agent keys found
Attempting OIDC authentication with <vault server>

Complete the authentication at:
    https://cilogon.org/device/?user_code=<code>
No web open command defined, please copy/paste the above to any web browser
Waiting for response in web browser
^C
```

[...]

PR #99, opened 2022-08-30
Make sure we set _condor_CREDD_HOST if we know it

shreyb:
This is a quick fix for Issue #98.  We simply set the environment variable `_condor_CREDD_HOST` to whatever schedd is passed in on the command line, if it is indeed passed in.  Verified that this fixes the issue as seen in #98 with `jobsub_transfer_data`.

closes #98

PR #100, opened 2022-08-30
set credd in jobsub_cmd wrapper.

marcmengel:
make sure schedd from either -name schedd or @schedd ends up as _condor_CREDD_HOST
fixes issue #98


marcmengel:
Merged in #99 code, to check if environment variable is set already.

PR #101, opened 2022-08-30
Addresses Issues 49 and 5

goodenou:
This PR adds parameters --offsite and --onsite to jobsub_submit command. These define the usage_model as 'OFFSITE' and 'OPPORTUNISTIC,DEDICATED', respectively. They are mutually exclusive parameters, so an error occurs if a user tries to use both.

If neither --offsite or --onsite is specified, the default usage_model is 'OPPORTUNISTIC,DEDICATED,OFFSITE'.

The appropriate usage_model is added to the .cmd file as +DESIRED_usage_model = "..." Any command-line definition of "--resource-provides=usage_model=..." overrides the default or what is set by the --offsite and --onsite parameters.

Additionally, if the usage_model contains 'ONSITE', "Fermigrid" is added to the list of DESIRED_SITES in the .cmd file.


goodenou:
Please have precommit.ci fix the issues.

lib/get_parser.py
```diff
@@ -324,7 +324,23 @@ def get_parser() -> argparse.ArgumentParser:
...
+        action="store_const",
+        const="OPPORTUNISTIC,DEDICATED",
+        default="OPPORTUNISTIC,DEDICATED,OFFSITE",
+        help="run jobs locally only; " "usage_model=OPPORTUNISTIC,DEDICATED",
+    )
+    parser.add_argument(
+        "--offsite",
+        dest="usage_model",
+        action="store_const",
+        const="OPPORTUNISTIC,DEDICATED,OFFSITE",
```
shreyb, line 339: Shouldn't this just be "OFFSITE" if we say `--offsite`?

lib/get_parser.py
```diff
@@ -324,7 +324,23 @@ def get_parser() -> argparse.ArgumentParser:
         action="store_true",
         help="Use jobsubdevgpvm01 etc. to submit",
     )
-
+    parser.add_argument(
```
shreyb, line 327: OK - this is a REALLY cool way to work the two flags in.

lib/get_parser.py
```diff
@@ -324,7 +324,23 @@ def get_parser() -> argparse.ArgumentParser:
...
+    )
+    parser.add_argument(
+        "--offsite",
+        dest="usage_model",
+        action="store_const",
+        const="OPPORTUNISTIC,DEDICATED,OFFSITE",
+        default="OPPORTUNISTIC,DEDICATED,OFFSITE",
+        help="run jobs offsite if slots slots are available; "
+        "usage_model=OPPORTUNISTIC,DEDICATED,OFFSITE",
+    )
```
shreyb, line 343: What do you think about making the two flags `--onsite` and `--offsite` mutually exclusive?  I know argparse has something like `mutually_exclusive_group` to do that if you think it's a good idea.

lib/utils.py
```diff
@@ -64,7 +64,7 @@ def set_extras_n_fix_units(
     #
     # outbase needs to be an area shared with schedd servers.
     #
-    if args["debug"] > 1:
+    if args["debug"]:
```
shreyb, line 67: I think this is actually supposed to be >1, since args.debug is actually a number, and Marc's idea was to make sure that we output that message only if something like `--debug 2` was entered.  Do you think this should not be the case, and that `--debug` should just be on or off (i.e. a `bool`)?

lib/utils.py
```diff
@@ -82,14 +82,13 @@ def set_extras_n_fix_units(
...
     args["kerberos_principal"] = get_principal()
-    args["usage_model"] = "ONSITE"
     args["uid"] = str(os.getuid())
 
     if not "uuid" in args:
         args["uuid"] = str(uuid.uuid4())
     if not "date" in args:
         args["date"] = datetime.datetime.now().strftime("%Y_%m_%d_%H%M%S")
-    if args["debug"] > 1:
+    if args["debug"]:
```
shreyb, line 91: Same comment as above

lib/utils.py
```diff
@@ -108,6 +107,21 @@ def set_extras_n_fix_units(
 
     args["resource_provides_quoted"] = [fixquote(x) for x in args["resource_provides"]]
 
+    # if the user defined the usage_model on the command line,
+    # we need to use their definition of usage_model, not ours
+    # also, we need to define the sites as 'Fermigrid' if they want to run locally
+    for r in args["resource_provides_quoted"]:
+        print(f"r is: {r}")
```
shreyb, line 114: I'm not sure what "r" is supposed to indicate to the user.  Perhaps say "resource_provides is {r}" or something like that (if I'm understanding correctly)?

lib/utils.py
```diff
@@ -108,6 +107,21 @@ def set_extras_n_fix_units(
...
+        if "usage_model" in r:
+            args["usage_model"] = ""
+        if r == 'usage_model="OPPORTUNISTIC,DEDICATED"':
+            args["site"] = "Fermigrid"
```
shreyb, line 118: In this case, don't we still also want to set `args["usage_model"]` to `OPPORTUNISTIC,DEDICATED`?  Or is the Fermigrid specification enough to steer the jobs to Fermigrid?

lib/utils.py
```diff
@@ -119,14 +133,14 @@ def set_extras_n_fix_units(
         os.makedirs(args["outdir"])
 
     # copy executable to submit dir so schedd can see it
-    if args["debug"] > 1:
+    if args["debug"]:
```
shreyb, line 136: Same as above

lib/utils.py
```diff
@@ -119,14 +133,14 @@ def set_extras_n_fix_units(
...
         sys.stderr.write(
             f"checking full_executable: {repr(args.get('full_executable', None))}\n"
         )
 
     if args.get("full_executable", False):
         dest = os.path.join(args["submitdir"], os.path.basename(args["executable"]))
-        if args["debug"] > 1:
+        if args["debug"]:
```
shreyb, line 143: Same as above

lib/utils.py
```diff
@@ -171,7 +185,7 @@ def set_extras_n_fix_units(
             e = f"{e}={v}"
         newe.append(e)
     args["environment"] = newe
-    if args["debug"] > 1:
+    if args["debug"]:
```
shreyb, line 188: Same as above

lib/utils.py
shreyb: From the submit file changes, I suspect that `Fermigrid` may be enough, but please confirm that.

shreyb (changes requested):
Thanks so much, Lisa, for writing that summary up - it made following the changes here much easier.  It looks mostly good, but I have a few questions I've raised here, and some comments that are definitely not blockers, but more solicitations for your input.

lib/get_parser.py
goodenou: Sorry, my misunderstanding of the request in Issue 49. I will change this.

lib/get_parser.py
goodenou: This makes sense. I will look into it. I was holding the user to a higher level of responsibility that you are are!

lib/utils.py
goodenou: I did not change these. I think my merge reverted these debug statements back to a previous version. I will fix them all.

lib/utils.py
goodenou: see above

lib/utils.py
goodenou: That print statement shouldn't be there. That was for my testing. I will remove it.

lib/utils.py
goodenou: Will fix this.

lib/utils.py
goodenou: Will fix this.

lib/utils.py
goodenou: Will fix this.

lib/utils.py
goodenou: There are two arg parameters, "--usage_model" and "--resource_provides", both of which could have a usage_model definition in them. (Note that the 'resource_provides' parameter could have OTHER definitions as well.) The 'resource_provides' parameter is left empty if the user doesn't define it. 'usage_model' is filled with the default value usage_model="OPPORTUNISTIC,DEDICATED,OFFSITE" if the user doesn't define it through the --offsite or --onsite parameters.

If someone sets the 'resource_provides' parameter to a usage_model in the command line, such as in "--resource-provides=usage_model=OPPORTUNISTIC,DEDICATED", we don't want the code to override their choice of usage_model. That is why I set the parameter "usage_model" to the empty string in line 116. This only happens if they have explicitly defined a usage_model in their command line.

Hopefully this makes sense.

lib/utils.py
shreyb: OK.  Yeah, there has been a lot of activity in the past few days.

lib/get_parser.py
goodenou: done

lib/utils.py
goodenou: done

lib/utils.py
goodenou: done

lib/utils.py
goodenou: done

lib/utils.py
goodenou: done

lib/utils.py
goodenou: done

lib/utils.py
goodenou: done

lib/utils.py
goodenou: That is correct.

lib/utils.py
```diff
@@ -108,6 +107,21 @@ def set_extras_n_fix_units(
...
     args["resource_provides_quoted"] = [fixquote(x) for x in args["resource_provides"]]
 
+    # if the user defined the usage_model on the command line,
+    # we need to use their definition of usage_model, not ours
+    # also, we need to define the sites as 'Fermigrid' if they want to run locally
+    for r in args["resource_provides_quoted"]:
+        print(f"r is: {r}")
+        if "usage_model" in r:
+            args["usage_model"] = ""
+        if r == 'usage_model="OPPORTUNISTIC,DEDICATED"':
```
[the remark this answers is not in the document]
goodenou: OK. This makes sense. I will make the relevant changes.

lib/get_parser.py
goodenou: done

lib/utils.py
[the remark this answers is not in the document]
goodenou: done

lib/utils.py
```diff
@@ -108,6 +107,26 @@ def set_extras_n_fix_units(
...
+    add_site = ""
+    for r in args["resource_provides_quoted"]:
+        if "usage_model" in r:
+            args["usage_model"] = ""
+            if "OFFSITE" not in r:
+                add_site = "Fermigrid"
+
+    # if the user chooses 'onsite' from the runtime params
+    # we need to define the sites as 'Fermigrid'
+    if args["usage_model"] is not "" and "OFFSITE" not in args["usage_model"]:
```
shreyb, line 122: So I had no idea that `is not` worked with strings, but when I did some investigation, it looks like the python interpreter tries to warn that it's improper.  Would you mind change that to either `args["usage_model"] != ""` or `(not args["usage_model"])`?

shreyb (changes requested):
One small change request in this review - just so in case future python versions get stricter, we don't get bitten by it.  Otherwise, this is great.  Have you run the unit tests on these changes?

shreyb (approved):

shreyb:
Will run unit tests and pre-commit.ci autofix

shreyb:
pre-commit.ci autofix

shreyb (approved):

lib/utils.py
```diff
@@ -108,6 +107,26 @@ def set_extras_n_fix_units(
...
+            args["usage_model"] = ""
+            if "OFFSITE" not in r:
+                add_site = "Fermigrid"
+
+    # if the user chooses 'onsite' from the runtime params
+    # we need to define the sites as 'Fermigrid'
+    if args["usage_model"] != "" and "OFFSITE" not in args["usage_model"]:
+        add_site = "Fermigrid"
+
+    if args["site"]:
```
shreyb, line 125: So I missed this in my last review, but thankfully our unit tests caught this!  This line will fail with a `KeyError` if `--site` is not set by the user.  What this line needs to be is something that assumes that `args["site"]` may not be set at all.  So either 
```python
if args.get("site", None):
``` 
or
```python
try:
    if args["site"]:
    args["site"] += ", " + add_site
except KeyError:
    args["site"] = add_site
```
or something like that.

lib/utils.py
```diff
@@ -108,6 +107,26 @@ def set_extras_n_fix_units(
 
     args["resource_provides_quoted"] = [fixquote(x) for x in args["resource_provides"]]
 
+    # if the user defined the usage_model on the command line,
+    # we need to use their definition of usage_model, not ours
+    # also, we define the site as 'Fermigrid' if they have not requested OFFSITE
+    add_site = ""
+    for r in args["resource_provides_quoted"]:
+        if "usage_model" in r:
```
shreyb, line 115: This line actually uncovered a mistake in our unit tests.  I'll open a PR for that, but it shouldn't conflict with your code.

shreyb (changes requested):

lib/utils.py
goodenou: fixed

lib/utils.py
goodenou: fixed

lib/utils.py
goodenou: ok

shreyb (commented):
Looks good.  After the token issue is fixed, I'll run it through the unit tests again and approve unless something else is brought up.

Thanks!

shreyb:
All unit tests pass except for one - the test_check_all_test_args test in `test_get_parser_unit.py`.  This test basically makes sure that our expected arguments get parsed correctly by things like dagnabbit.py.   Could you please add `"--offsite"` and `"--onsite-only"` to the list of args in the fixture `all_test_args` in that file?  Here's a link to the line number you'd need to add the flags at:  https://github.com/marcmengel/jobsub_lite/blob/master/tests/test_get_parser_unit.py#L138  

Assuming nothing else that we haven't caught, that should be it for this PR.

tests/test_get_parser_unit.py
```diff
@@ -156,6 +156,8 @@ def all_test_args():
         "--use-pnfs-dropbox",
         "--verbose",
         "--devserver",
+        "--onsite-only",
+        "--offsite",
```
shreyb, line 160: Agh, I forgot one last thing.  My mistake.  Down in test_get_parser_all, please add _either_ one of `"--onsite-only"` or `"--offsite"` to the `args_exclude_list` in the `test_get_parser_all` function.  

When these unit tests were first written, there were no mutually exclusive argument groups.  When I added the first one for singularity stuff, I had to add that list (and exclude any arg in that list from the test) to make the unit test work.

Thanks!

shreyb: The requested change should be made on line 225, as I see it here.

shreyb (changes requested):
Sorry Lisa - one last thing I forgot.  But with that change, all unit tests pass, which is great!

shreyb (approved):

shreyb:
Thanks Lisa.  Merging this now.

PR #103, opened 2022-08-31
decode_token fix

marcmengel:
fix for underscores/slashes stolen from httokendecode.


shreyb (approved):
Tested on fermilab, mu2e, and dune tokens and it works.

PR #104, opened 2022-08-31
Added fixture setup_ups to set up UPS in test environment

shreyb:
This PR just adds a fixture to the `test_packages_unit` tests that sets up UPS, in case it's not already set up in the user's environment.  In a standard terminal/shell, a developer can just set up UPS/Spack, and then run the pytest command manually; but in an IDE like VSCode, PyCharm, etc., the environment for the native IDE testing functionality is often separate from the built-in terminal, and so simply sourcing a script won't work.  This change should ensure that no matter where a developer is running the tests from, as long as UPS is available, the unit tests should behave as expected.

### DIFF
--- a/tests/test_creds_unit.py
+++ b/tests/test_creds_unit.py
@@ -20,7 +20,7 @@ from test_unit import TestUnit
 @pytest.fixture
 def needs_credentials():
     os.environ["GROUP"] = TestUnit.test_group
-    return creds.get_creds()
+    return creds.get_creds({"role": "Analysis"})
 
 
 @pytest.fixture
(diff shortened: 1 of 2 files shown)


shreyb:
Notes:
1.  I verified that with this change, all unit tests now pass
2. Requested @marcmengel as reviewer because he wrote the original unit tests, and is the most familiar with the various package managers we use like UPS and Spack.

marcmengel (changes requested):
Looks good overall, should also try /grid/fermiapp/products/common/etc/setups.sh, in case
we have /grid/fermiapp mounted but not /cvmfs (i.e. fifeutilgpvm01) .
maybe check which one exists in setup_ups.

shreyb:
Good point, Marc.  I added in a check for the ups setup script in /grid/fermiapp if CVMFS fails.  Please let me know if this is OK and if there are any other issues you see with the PR.

shreyb:
@marcmengel does this look good now?

PR #108, opened 2022-09-01
Decode token tests

marcmengel:
Tests for decode_token.sh.   Included data files which are hand-hacked files that look enough like actual tokens to test the utility, but do not have valid signatures, nor do they contain actual usernames or token-id's.   They are based on actual tokens, but have been obfuscated.


tests/test_decode_token_unit.py
```diff
@@ -0,0 +1,77 @@
...
+    assert int(lines[0]) > time.time()
+
+
+def test_decode_token_ext_group():
+    """make sure we can extract wlcg.groups and get whole list"""
+    lines = do_decode_token_sh("decode_token_tests/dp1", "wlcg.groups")
+    assert lines[0] == '["/dune","/dune/production"]\n'
+
+
+def test_decode_token_past_bugs(token):
```
shreyb, line 69: One nitpicky thing here:  Does this actually use the `token` fixture?  It looks like this test goes through all the sample tokens and runs checks on them, not a new token.  If that's the case, we should probably not pass `token` as an argument here.

shreyb (changes requested):
Looks good, except for the one comment I had.  If you agree, please change that, and then I'm good to approve this.

marcmengel:
Since I made the change you requested Shreyas, I'm going ahead and merging this without waiting for you to approve.

Issue #110, opened 2022-09-02
jobsub commands hang when user needs to authenticate


retzkek:
Running any jobsub or condor commands ~that point to jobsub_cmd~ (edit: every command actually) now seem to hang when the user doesn't have a valid token locally or in vault and needs to authenticate in the browser. It's not actually hanging, we're just not displaying the message that tells them to authenticate any more (#80).

https://github.com/marcmengel/jobsub_lite/blob/02c65037ef3096c3a3335136da749626d0a2d08c/lib/fake_ifdh.py#L123

Related, jobsub_cmd should have a --debug option (which it would get when changed to use `get_base_parser()` added in #73) and should pass that to the `get{Token,Proxy}()` functions.

shreyb:
I'm going to split out the "related" comment, since the bug described in this issue is pretty urgent.

shreyb:
Steps to reproduce:

1. Destroy all vault tokens `rm /tmp/vt*`
2. I happened to know I've never tried to submit jobs as the `gm2` experiment, so use them
3. Try jobsub_submit with gm2 as experiment:

```
# Works because --debug is turned on:
$ ./jobsub_lite/bin/jobsub_submit -G gm2 --resource-provides=usage_model=DEDICATED,OPPORTUNISTIC,OFFSITE --devserver  --debug 1  file:///usr/bin/printenv

Couldn't find a valid proxy.

PCI extension info:
 Path length: -1
 Policy language not specified.
 Policy file not specified.
Number of bits in key :2048
Files being used:
 CA certificate file: none
 Trusted certificates directory : /etc/grid-security/certificates
 Proxy certificate file : /tmp/x509up_gm2_Analysis_10610
 User certificate file: /tmp/x509up_u10610
 User key file: /tmp/x509up_u10610
Output to /tmp/x509up_gm2_Analysis_10610
Your identity: /DC=org/DC=cilogon/C=US/O=Fermi National Accelerator Laboratory/OU=People/CN=Shreyas Bhat/CN=UID:sbhat
Using configuration file /home/sbhat/.glite/vomses
Using configuration file /etc/vomses
Contacting  voms1.fnal.gov:15001 [/DC=org/DC=incommon/C=US/ST=Illinois/O=Fermi Research Alliance/OU=Fermilab/CN=voms1.fnal.gov] "fermilab" Done
Creating proxy to /tmp/x509up_gm2_Analysis_10610 ....................................................+++
.......................+++

[...]

PR #112, opened 2022-09-02
Bugfix for htgettoken URL not showing up

shreyb:
This just makes sure to print the stdout from htgettoken, since under normal conditions, it's not that much output.

More importantly, the user needs to see the authentication URL if they're getting the token for the first time

### DIFF
--- a/lib/fake_ifdh.py
+++ b/lib/fake_ifdh.py
@@ -116,11 +116,10 @@ def getToken(role: str = DEFAULT_ROLE, debug: int = 0) -> str:
             cmd = f"{cmd} -r {role.lower()}"  # Token-world wants all-lower
 
         if debug > 0:
-            # send htgettoken output to stderr because invokers read stdout
             sys.stderr.write(f"Running: {cmd}")
-            cmd += " >&2"
-        else:
-            cmd += " >/dev/null 2>&1"
+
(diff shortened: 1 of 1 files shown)


shreyb:
Closes #110

marcmengel (commented):
This says send output to stderr in the comments but does the opposite... 1>&2 is sending stdout to stderr...

shreyb:
Thanks for catching that @marcmengel .  It's now fixed.

retzkek:
Wait, no shouldn't the code have been changed to match the comment? Or maybe I don't understand, so why do we want to redirect either stdout or stderr instead of passing them both through respectively?

shreyb:
The code as written was doing the right thing in terms of passing stderr to stdout.  The comment in the code however, was wrong, and said that output was getting passed to stderr.  So this last commit was just changing that comment so that it matched what the code was doing.  We want everything to go to stdout, since `os.system()` sends all stdout to the interpreter's (user's) stdout.

shreyb:
Talked to @retzkek offline.  He's right that `os.system()` already dumps everything to the interpreter's stdout, so there's no reason for us to do so.  I changed the code to reflect this.

retzkek (approved):
LGTM

Issue #113, opened 2022-09-07
RCDS Failover needed


marcmengel:
SO at the moment I'm writing this, rcds01 is reachable but rcds02 isn't.   Our tarball-upload code currently picks one or the other at random, but having picked rcds02, retries until it hits a timeout and never tries rcds01. 
I don't think we can switch hosts once we actually uploaded a tarfile, (because the other server won't know about it) 
 but we could have a sanity check up front to see if the host is alive before picking the host we're using...

marcmengel:
So the code was trying to do this, but was replacing the {dropbox_server} in self.pubapi_base_url_formatter the
first time in its wrapper function and then never trying the new server it was getting from its iterator.

I'm trying a new variable pubapi_base_url_formatter_full which never gets overwritten and is formatted
into pubapi_base_url_formatter on each retry.

marcmengel:
Fixed in #114

PR #114, opened 2022-09-07
first draft at fixing rcds failover

marcmengel:
Fix for #113 -- try alternate host on retries.


marcmengel:
I think this code was lifted from jobsub, so we probably have the same bug there.

shreyb (approved):
Looks good, as long as it works.  My guess is that it fixes the issue because the formatter object isn't mutable?

marcmengel:
The old code replaced the {dropbox_server} in the URL format the first time, and then never thereafter.
it kept getting new names from the iterator, but never actually used them 'cause the key was no longer
in the format.

Issue #109, opened 2022-09-02
local submit directories/sandbox transfer directories and cleanup


marcmengel:
Currently jobsub_lite creates directories under $HOME/.jobsub_lite, with submit files, and is the place where condor_transfer_data returns results; which works, but repeated use will eventually run the user out of quota 
in their $HOME area.

As we move to jobsub_fetchlog, this becomes increasingly invisible to the users, so it will not occur to
them to clean this up. 

We should revisit this, and pick a "standard" location, and make sure files get cleaned up eventually.

Email disucussion included this from Kevin:

After some more thought, while /run/user ($XDG_RUNTIME_DIR) would be convenient, our use would be against spec so we’d need to be careful, especially if dumping logs in there (true of any tmpfs solution). Interactive nodes give /run a few GB and generally have plenty of RAM (depending on what users are doing of course) so maybe it’d be OK…
 
tmpfs                              5.8G  3.0M  5.8G   1% /run
 
$XDG_CACHE_HOME (default $HOME/.cache) or $XDG_STATE_HOME (default $HOME/.local/state) are probably more appropriate locations, but then we’d need to do cleanup. Maybe that’s not so bad - clean up as we go, and maybe also have a routine run with every jobsub command that looks for old submission dirs that didn’t get cleaned up.
 
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables

marcmengel:
Earlier comment from Kevin:
  

> Maybe we should just delete the local sandbox after successful submission? Fetchlog will re-create the directory if it doesn’t exist, and them I suppose it should delete the directory as well when done. Or just put them in some tmpfs location (/run/user, /run/jobsub?) so we get OS cleanup occasionally.

 
I think this would work fine -- **if** we make sure to include the submit file, etc in the files we transfer to the
job, so we get it back at the end,  Otherwise the submit file disappears, which could make debugging difficult.

marcmengel:
So after discussion I think we're looking at:
1) making sure all the job files (i.e. the simple.cmd, etc.) get transferred to and from the job
2) cleaning out the sandbox directory after submission
3) Copying submit files to current directory, and or  a subdirectory of current directory if --no-submit 
4) cleaning out the sandbox directory after condor_transfer_data and copying elsewhere in jobsub_fetchlog
We could also have jobsub_submit and jobsub_fetchlog clean up any more-than-a-week-old submit
[...]

PR #115, opened 2022-09-07
Sandbox directory cleanup

marcmengel:
This addresses  #109.  We now include more files in the transfer_{input,output}_data files so we can just remove
the local sandbox directory after submitting, and again after running condor_transfer_data in jobsub_fetchlog.

The directory is now in $HOME/.cache/jobsub_lite by default.


marcmengel:
Hm.... dataset dag jobs are going held:
HoldReason = "Error from slot1_63@fnpc7019.fnal.gov: STARTER at 131.225.167.183 failed to send file(s) to <131.225.240.23:9615>: error reading from /storage/local/data1/condor/execute/dir_37014/simple.sh: (errno 2) No such file or directory; SHADOW failed to receive file(s) from <131.225.167.183:46880>"
So not getting wrapper script file back from worker to dagman area?  Need to look into this further...

marcmengel:
Okay, so I was confused by condor renaming the script to condor_exec.exe .  I now have the scripts linking themselves back to their original names.

bin/jobsub_submit
```diff
@@ -235,6 +229,11 @@ def main():
             os.chdir(varg["submitdir"])
             submit(os.path.join(submitdir, "*.cmd"), varg, schedd_name)
 
+    if varg.get("no_submit", False):
+        print(f"Submission files are in: {varg['submitdir']}")
```
shreyb, line 233: Love that - very helpful to the user.

lib/utils.py
```diff
@@ -24,9 +24,30 @@
...
+def cleanup(varg: Dict[str, Any]) -> None:
+    """cleanup submit directory etc."""
+    os.chdir(f'{varg["submitdir"]}/..')
+    cleandir(varg["submitdir"])
+    # now clean up old submit directories that weren't
+    # cleaned up by jobsub_submit at the time
+    with os.scandir(".") as it:
+        for entry in it:
+            sb = os.stat(entry.name)
+            if entry.name.startswith("js_") and sb.st_mtime - time.time() > 604800:
```
shreyb, line 47: Shouldn't this subtraction be reversed?  That is `time.time() - sb.st_mtime`?

shreyb (changes requested):
All good except for the one comment I had about calculating how old a submit dir is.

lib/utils.py
marcmengel: Ack!  yes

marcmengel: Fixed.  I think I'm getting back all the files I should now for both dag and plain submits, so I think this is ready to merge.

shreyb (approved):

shreyb:
Just to make sure, are all the changes for this done, or are you still figuring out some of the transfer_{input,output}_files stuff?

PR #116, opened 2022-09-08
cleanup tests

marcmengel:
Should have been in #115, but there you go...


marcmengel:
only tests, just rolling it in.

Issue #106, opened 2022-09-01
Remove any hard-coded hostnames from code, put into separate config that is version-controlled on private  git


shreyb:
At yesterday's jobsub project meeting, we discussed how much information is reasonable to have on a public repository.  We decided that as much as possible, we should try to remove hostnames from this repository, even if there's no actual security threat.  

The current idea is to keep the configuration in a separate RPM, version-controlled on a private git repository.  Then, when jobsub_lite is deployed, both the jobsub_lite RPM and the config rpm should be installed.

shreyb:
Assigned to @marcmengel  since he's started work on it. Marc - feel free to unassign the issue if you need to step away from it.

marcmengel:
So I have a pull request for phase 1 of this; which concentrates hostnames into config.d and spec subirectories.

I have made a new repo https://github.com/marcmengel/jobsub_lite_config with just those config.d and spec subdirectories.

Next is to actually remove the config.d and spec subdirectories from the main repo... the question is do we just remove them at the HEAD or do we do git filter-branch evil to clean them out throughout history and clean out the past references...

shreyb:
This is completed now, right?

shreyb:
This is completed, right?  After talking with FNAL cybersecurity, they said we didn't have to git-filter-branch anything.

marcmengel:
Yes, I believe we're done.

PR #118, opened 2022-09-09
Host cleanup part 1

marcmengel:
This pull request moves all literal hostnames  into the config.d and spec subdirectories, so they can be
split out into a separate, private, Git repo.  Two of them, in the test code, are replaced by obfuscating
functions, which should still hide them from things that scan for hostnames, and will also give wrong/different
answers offsite.


marcmengel:
This is a first step to resolving #106

shreyb (approved):
LGTM.

PR #120, opened 2022-09-13
using jobsub_lite_config as submodule, cleaning out directories moved there

marcmengel:
This is part 2 of cleaning up hostnames.   We may still do a part 3 to clean up history, but that will be a change that makes everyone recheck-out the whole repo, 'cause we're going to rewrite history there.


shreyb (approved):

shreyb:
LGTM

Issue #119, opened 2022-09-12
Document list of Environment Variables we clear in #84


shreyb:
In #84, we implemented the automatic (but opt-out-able) clearing of certain environment variables.  These need to be documented in the FIFE Mediawiki at fifewiki.fnal.gov.

marcmengel:
It is at: https://fifewiki.fnal.gov/wiki/Jobsub_Wrapper_Scripts#Startup_actions

Although, I think we did **not** make it opt-out-able...

shreyb:
Do you want to do that as a part of this issue, or should I split it out?

marcmengel:
So the opt-out flag is added, as are the docs.

PR #121, opened 2022-09-14
make environment cleanup optional

marcmengel:
last tidbit for #119, #84 adding a command line option to not clear our list of bad environment variables in the wrapper.


shreyb (approved):
LGTM

Issue #123, opened 2022-09-14
jobsub_q prints errors


rlcee:
jobsub_q prints errors (at least for rlc, on fifeutilgpvm01 with --devserver )
[rlc@fifeutilgpvm01 ~]$ rpm -q -a | grep jobsub
jobsub_dropbox_cleanup-1.2.4-1.x86_64
jobsub_lite-beta13-1.el7.x86_64

```
-- Failed to fetch ads from: <131.225.240.53:9615?addrs=131.225.240.53-9615&alias=jobsub03.fnal.gov&noUDP&sock=schedd_2728_c273> : jobsub03.fnal.gov
AUTHENTICATE:1003:Failed to authenticate with any method

-- Failed to fetch ads from: <131.225.161.104:9615?addrs=131.225.161.104-9615&alias=jobsub02.fnal.gov&noUDP&sock=schedd_3341_3a01> : jobsub02.fnal.gov
AUTHENTICATE:1003:Failed to authenticate with any method

-- Failed to fetch ads from: <131.225.161.91:9618?addrs=131.225.161.91-9618&alias=gpce03.fnal.gov&noUDP&sock=schedd_3397_a34f> : gpce03.fnal.gov
AUTHENTICATE:1003:Failed to authenticate with any method

```

marcmengel:
This is because we have several schedd nodes (i.e. jobsub02, jobsub03) that aren't configured to 
accept tokens (yet). These errors will go away once we roll token support out more completely.

Also, jobsub_q doesn't accept "--devserver" even in the current (jobsub_client) code. 
We *could* make the new one accept it... but you can do:
`jobsub_q -name jobsubdevpgmv01.fnal.gov`
(in jobsub_lite) to see just that schedd.

marcmengel:
Or rather` jobsub_q -G group -name jobsubdevgpvm01.fnal.gov`  -- you still need -G

Issue #122, opened 2022-09-14
how jobsub_q handles missing -G


rlcee:
- if -G is not set, traceback is not expected
- old jobsub_q gives a reasonable response if -G is not set.

```
[rlc@fifeutilgpvm01 ~]$ jobsub_q
Traceback (most recent call last):
  File "/opt/jobsub_lite/bin/jobsub_q", line 114, in <module>
    main()
  File "/opt/jobsub_lite/bin/jobsub_q", line 64, in main
    f"{sys.argv[0]} needs -G group or $GROUP in the environment."
AttributeError: /opt/jobsub_lite/bin/jobsub_q needs -G group or $GROUP in the environment.

```

marcmengel:
This was resolved by making the -G error raise SystemExit  instead of  KeyError; SystemExit does not print a stack trace.

PR #124, opened 2022-09-15
Arg handling and help fixup

marcmengel:
This pull request f1xes #122, and parts of #123, and a few other bits brought up by users, and improves
command-line backwards compatability on the non jobsub-submit wrappers.

You can now use --jobid=1234.0@schedd or 1234.0@schedd or -name schedd 1234.0 
interchangeably on all the commands that take jobid's. 

You can do any of jobsub_q --better-analyze or jobsub_q -better-analyze or condor_q -better-analyze,
ditto for -long / --long.

It also includes the condor_q --help info in jobsub_q --help, etc.

### DIFF
--- a/lib/get_parser.py
+++ b/lib/get_parser.py
@@ -44,10 +44,11 @@ class StoreGroupinEnvironment(argparse.Action):
         setattr(namespace, self.dest, values)
 
 
-def get_base_parser() -> argparse.ArgumentParser:
+def get_base_parser(with_jobid: bool = False) -> argparse.ArgumentParser:
     """build the general jobsub command argument parser and return it"""
     parser = argparse.ArgumentParser()
     group = parser.add_argument_group("general arguments")
+
     group.add_argument(
         "-G",
         "--group",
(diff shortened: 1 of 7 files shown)


marcmengel:
...except  jobsub_fetchlog won't take `-name schedd 1234.0`.   Not sure if I want to go back and change that.

shreyb:
I agree about your `jobsub_fetchlog` point.  We probably don't want that.

lib/get_parser.py
```diff
@@ -44,10 +46,20 @@ def __call__(
         setattr(namespace, self.dest, values)
 
 
-def get_base_parser() -> argparse.ArgumentParser:
+def get_base_parser(
```
shreyb, line 49: I don't love the idea of adding arguments to the `get_base_parser` function as needed, just because that could turn into a mess if we extract it out to many kinds of parsers.  What would you think about having the base parser stay the same, that is, the common set of flags that every binary needs, and then having different module-level functions in this file that the binaries could call?  So something like:

```python
def get_base_parser():
    same_args_as_now()
    return parser

def get_jobid_parser():
    base_parser = get_base_parser()
    parser.add_argument("-J", "--jobid"....)
    return parser
```

and so on?

shreyb (changes requested):
It all looks good, but I'm not a big fan of expanding `get_base_parser`.  Please take a look at the comment and see if you agree.

lib/get_parser.py
marcmengel: Okay.  That sounds reasonable.  I've pushed that change.

shreyb (approved):
Looks good!

PR #130, opened 2022-09-21
Tracing revisited

marcmengel:
Split this back out into a fresh pull request, since I don't see how to reopen the old one.

Issues:
  * supporting needed RPMs / python libraries
  * providing hook for allowing  tags/attributes, the function decorator can't do that.


marcmengel:
Now github says this is the same as master, but I have diffs on my end.. sigh.

PR #131, opened 2022-09-21
Tracing

marcmengel:
Picking up tracing again here.

The open issues are:
 * providing a way to do tags and attributes
 * support / maint  of 9(!) RPMs needed for opentelemetry,etc.


marcmengel:
Sorry.  This is on hold pending adding more trace info, etc. 

I'll mark it a draft.

marcmengel:
Okay, so I've added more tracing, exposed a start_as_current_span method you can use with "with start_as_current_span", and an "add_event" method so you can do more detailed logging, and the code
now makes fairly detailed logs of submits (not so much of other wrapped condor commands, although we log them, too). See https://landscape.fnal.gov/jaeger/trace/1cbcb9e6bed298853426e96d2ffc2fe3 for a sample soup-ed up  submit trace

marcmengel:
Also over in the jobsub_lite_config repo, we have the spec file for the all-in-one python3 yaeger tracing dependencies for SL7, which provides all 11 or so python packages in one handy RPM.  Since Centos9 includes these RPMs already, this should go away eventually with SL7.

marcmengel:
This is rebased off master again, and ready for retesting.

retzkek:
The implementation looks OK to me. One request though, perhaps as another PR: could we inject the trace ID into the job classad? `JOBSUB_TRACE_ID` or something like that. Then we could use it in IFDH etc and really live the dream!

marcmengel:
Should we call it TRACEPARENT and also pass it in the environment?   Then other OTEL pacakges should pick it up, I think? 
https://github.com/open-telemetry/opentelemetry-specification/issues/740

shreyb:
@marcmengel is this ready for @retzkek to re-review?

marcmengel:
I thinks so.. I'll request him.

retzkek (approved):
LGTM

shreyb:
The conflicts for `bin/jobsub_submit` are substantial, mostly because a lot of that code was moved to `lib/submit_support.py`.  I'd prefer to wait until Marc comes back and then we can work on resolving the conflicts.

lib/tracing.py
```diff
@@ -0,0 +1,168 @@
...
+        return trace.get_current_span()
+
+    def get_propagator_carrier() -> Dict[str, str]:
+        carrier: Dict[str, str] = {}
+        TraceContextTextMapPropagator().inject(carrier)
+        return carrier
+
+except:
+    # if we can't import the stuff, here's a little mock so we don't crash
+    print("Note: tracing not available here.")
```
shreyb, line 77: Could you please add to this message that we'll still continue the operation without tracing?  The first time I saw this, I Ctrl-C'd because I thought it wouldn't work.

shreyb: I love the fact that it still DOES work due to your mock.  That's a great touch.

shreyb (changes requested):
One small change request.

shreyb (approved):
LGTM.  I was able to trace a jobsub_submit, get the traceID, and find it in jaeger.

shreyb:
Note that deploying this will require #463 to be completed.  But the code itself is good.

PR #135, opened 2022-09-28
--dag shouldn't take an argument, …

marcmengel:
…it should treat the filename/executable as the dag file


shreyb (approved):
LGTM.  Along with this, could you please update https://fifewiki.fnal.gov/wiki/Getting_started_with_jobsub_lite#Submit_DAG with a quick example of using `--dag` (not as a part of this PR, of course)?

PR #138, opened 2022-09-28
If X509_USER_PROXY is set, we assume the proxy should live there - fix for kherner issue 

shreyb:
If X509_USER_PROXY is set, we assume the proxy should live there or should be generated there


marcmengel (approved):
LGTM

Issue #134, opened 2022-09-28
Option `--tar_file_name dropbox:///path` doesn't seem to work properlky


vitodb:
In the submission I use
```
--tar_file_name dropbox:///home/vito/sbndcode_v09_59_00_e20_prof.tar --use-cvmfs-dropbox
```
the tarball is uploaded to RCDS, but jobs try to get it as local tarball.
In the stderr I get
```
...
 checking prefix / -> /
Exception:Path dropbox:///home/vito/sbndcode_v09_59_00_e20_prof.tar did not match in configuration
at:Tue Sep 27 21:57:41 2022

tar (child): .unwind_1/sbndcode_v09_59_00_e20_prof.tar: Cannot open: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
tar: Error is not recoverable: exiting now
...
```
and the simple.sh script has
```
# -f files for input


# --tar_file_name for input

  
    # RCDS unpacked tarfile
    
      INPUT_TAR_FILE=/cvmfs/fifeuser2.opensciencegrid.org/sw/sbnd/18bd649a28f4ea976629004c83334d341442d36ca802c7f93fe6a14154b24d30
      export INPUT_TAR_FILE
    
  

      # tarfile to transfer and unpack
    mkdir .unwind_1
    
    ${JSB_TMP}/ifdh.sh cp dropbox:///home/vito/sbndcode_v09_59_00_e20_prof.tar .unwind_1/sbndcode_v09_59_00_e20_prof.tar
    tar --directory .unwind_1 -xzvf .unwind_1/sbndcode_v09_59_00_e20_prof.tar
    
      INPUT_TAR_FILE_1=`pwd`/.unwind_1
      export INPUT_TAR_FILE_1
```

marcmengel:
Hm... that's weird.  The "dropbox:" string shouldn't be getting through to the simple.sh script at all, it should have been converted to a /cvmfs path...  
ah..., there we go.  It was putting the converted path *and* the dropbox: path in.. .sigh. 
```
         if tfn.startswith("dropbox:"):
            # move it to dropbox area, pretend they gave us plain path
            path = tarfile_in_dropbox(args, tfn[8:])
            if path:
                res.append(path)
            else:
                res.append(tfn.replace("dropbox:", ""))

        res.append(tfn)
```
in the `if path:` bit it should be setting tfn to the value, not appending the value, so the append following will put it in the result.

[...]

PR #144, opened 2022-10-06
use INPUT_TAR_DIR_LOCAL for rcds unpacked tar directories, 

marcmengel:
…and wait for the first one to show up

Second pull request on #134 

"This time Rocky, for sure!!!"

### DIFF
--- a/templates/simple/simple.sh
+++ b/templates/simple/simple.sh
@@ -215,11 +215,24 @@ ${JSB_TMP}/ifdh.sh cp -D {{fname}} ${CONDOR_DIR_INPUT}
   {%if tfname[:6] == "/cvmfs" and tfname[-7:] != ".tar.gz" %}
     # RCDS unpacked tarfile
     {%if loop.first%}
-      INPUT_TAR_FILE={{tfname}}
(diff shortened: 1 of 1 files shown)


shreyb (commented):
Question here:  As a part of this change, are we breaking backward compatibility - in that INPUT_TAR_FILE will no longer work with RCDS?

marcmengel:
Hm.. I thought INPUT_TAR_FILE already didn't work with RCDS...I mean, the tarfile isn't there, just the contents...

marcmengel:
Yes, in the current jobsub code:
https://cdcvs.fnal.gov/redmine/projects/jobsub/repository/revisions/master/entry/lib/groupsettings/JobSettings.py#L1035
if we have a "cid=" in our tar_file_name we set INPUT_TAR_DIR and INPUT_TAR_DIR_LOCAL, otherwise we set INPUT_TAR_FILE and unpack said tarfile.  We don't set both together.

shreyb:
Ah!  Got it.  Great - then this is good to go.

shreyb (approved):

PR #145, opened 2022-10-11
get username from production token

marcmengel:
One way to make production job submission as regular user transparent...


marcmengel:
This needs testing... so we shouldn't merge it yet..

shreyb:
I missed your comment, Marc.  Can you please move this back to draft status until you're satisfied with the testing?

marcmengel:
Closing this because #248 means we don't really need to set $USER anymore, so getting it from a token isn't needed.

PR #146, opened 2022-10-13
Jobsub q format fix

marcmengel:
So after discussing it at the meeting yesterday, I thought "How hard could it be to make the output more like old jobsub_q?"
so here it goes...
The main code involved is the logic in turning it on and off, and a few bits of classadd expression here and there.

### DIFF
--- a/bin/jobsub_cmd
+++ b/bin/jobsub_cmd
@@ -152,7 +152,7 @@ def main() -> None:
                     'strcat(split(GlobalJobId,"#")[1],"@",split(GlobalJobId,"#")[0])',
                     "-format",
                     "%-10s\t",
-                    "Owner",
+                    """(DAGNodeName=!=''?strcat(" |-",DAGNodeName):Owner)""",
                     "-format",
                     "%-11s ",
                     'formatTime(QDate,"%m/%d %H:%M")',
@@ -169,11 +169,17 @@ def main() -> None:
                     "%6.1f ",
                     "ImageSize/1024.0",
                     "-format",
-                    " %s ",
+                    "%s",
                     "Cmd",
                     "-format",
-                    " %20.20s\n",
+                    " %-.20s",
+                    "Args",
+                    "-format",
+                    " %-.20s",
                     "Arguments",
+                    "-format",
+                    "\n",
+                    "Owner",
                 ]
             )
 
--- a/bin/jobsub_cmd
+++ b/bin/jobsub_cmd
@@ -148,11 +148,8 @@ def main() -> None:
             execargs.extend(
                 [
                     "-format",
-                    "%12s@",
-                    'split(GlobalJobId,"#")[1]',
-                    "-format",
-                    "%-25.25s\t",
-                    'split(GlobalJobId,"#")[0]',
+                    "%-40s",
+                    'strcat(split(GlobalJobId,"#")[1],"@",split(GlobalJobId,"#")[0])',
                     "-format",
                     "%-10s\t",
                     "Owner",
@@ -181,7 +178,7 @@ def main() -> None:
             )
 
             print(
-                "JOBSUBJOBID                     \tOWNER     \tSUBMITTED     RUNTIME"
+                "JOBSUBJOBID                             OWNER       \tSUBMITTED     RUNTIME"
                 "   ST PRIO   SIZE  COMMAND"
             )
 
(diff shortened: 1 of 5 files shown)


bin/jobsub_cmd
```diff
@@ -85,6 +87,33 @@ def main() -> None:
...
+        ]:
+            default_formatting = False
+
+        if (
+            i
+            in [
+                "-constraint",
+                "-factory",
+                "-autocluster",
+                "-constraint",
```
shreyb, line 109: I believe `-constraint` should not be in here twice.

bin/jobsub_cmd
```diff
@@ -85,6 +87,33 @@ def main() -> None:
...
+                "-unmatchable",
+                "-allusers",
+            ]
+            or i[0] in "0123456789"
```
shreyb, line 113: Is this for the formatting directives like `%0.1s` or whatever those are?

bin/jobsub_cmd
```diff
@@ -103,8 +132,62 @@ def main() -> None:
...
 
+        if default_constraint:
+            execargs.extend(
+                [
+                    "-constraint",
+                    f'''Jobsub_Group=="{os.environ['GROUP']}"''',
+                ]
+            )
+
+        if default_formatting:
```
shreyb, line 146: Do we want to actually default to the old jobsub format?  Or keep  the current default format, with the ability to get the old format?  That was my understanding from yesterday's meeting (I'm compiling those minutes now...sorry for the delay).  @retzkek  - how did you interpret this decision?

bin/jobsub_fetchlog
```diff
@@ -65,6 +65,7 @@ def main():
         help='format for downloaded archive: "tar" (default, compressed with gzip) or "zip"',
         default="tar",
     )
+    parser.add_argument("job_id", nargs="?", help="job/submission ID")
```
shreyb, line 68: This should be `--jobid`, like https://github.com/marcmengel/jobsub_lite/blob/master/lib/get_parser.py#L89 specifies.  Could this be the same as the line I linked above, just to standardize the arguments across the tools?

templates/simple/simple.cmd
```diff
@@ -36,6 +36,9 @@ when_to_transfer_output = ON_EXIT_OR_EVICT
 +JOB_EXPECTED_MAX_LIFETIME = {{expected_lifetime}}
 notify_user = {{email_to}}
 
+# set command to user executable for condor_q, etc.
++Cmd = {{executable|basename}}
+
```
shreyb, line 41: Ooh - I like this.

shreyb (changes requested):
Very nice - but I have a couple of change requests and questions.

bin/jobsub_cmd
marcmengel: no, that's for ClusterId's -- if they give us a 1234.0 or 1234.0@schedd that's a constraint for a single job, so we don't want to give the (jobsub_q default) group constraint.

shreyb: I see.  That makes sense.  Thanks for clearing that up for me.

bin/jobsub_cmd
marcmengel: Yes, I was going to bring this up for discussion again; Should we not default to the old behavior? If people are reading current jobsub output, say in a script, then we should keep the old output... do we think users are wrapping jobsub_q or parsing its output in other scripts?

shreyb: Resolved - jobsub working group decided to have default behavior match old jobsub.  So this is good.

bin/jobsub_fetchlog
marcmengel: So... we take *either*` --jobid 1234@schedd `*or* just `1234@schedd`   the former is "jobid", the latter is "job_id", and after we parse the arguments we set one from the other if it is not set.

bin/jobsub_cmd
marcmengel: fixed.

bin/jobsub_fetchlog
shreyb: I did miss that - thanks for pointing it out.  In that case, I'm good with this.

shreyb (approved):
Looks good - thanks!

PR #155, opened 2022-10-26
handle incomplete transfers

marcmengel:
Right now if you get an error in transfer_data(), jobsub_fetchlog just exits without tarring up the (partial) results.
Modified to log the error and warn the user the transfer may be incomplete, but still tars/delivers the results we
did get.


shreyb (approved):
LGTM

PR #156, opened 2022-10-26
get everyone in the group, constraint fix...

marcmengel:
Of course the moment I merged the jobsub_q changes, I saw a  problem... it was only showing *my* jobs by default, and the constraint wasn't working right. 

Fixed here.


shreyb (approved):
Looks good.

Issue #154, opened 2022-10-25
Improve error messages when invalid units provided


retzkek:
Jobsub is pretty strict about units in arguments, e.g. for memory defaulting to MiB unless there's a one of kb, mb, gb, or tb (case-insensitive) given as a suffix. The help text does specify this. But what if someone forgets and omits the "b"? 

```
$ jobsub_submit --memory=100M ...
Traceback (most recent call last):
  File "/opt/jobsub_lite/bin/jobsub_submit", line 237, in <module>
    main()
  File "/opt/jobsub_lite/bin/jobsub_submit", line 180, in main
    set_extras_n_fix_units(varg, schedd_name, proxy, token)
  File "/opt/jobsub_lite/lib/utils.py", line 197, in set_extras_n_fix_units
    fix_unit(args, "memory", memtable, -1, "b", -2)
  File "/opt/jobsub_lite/lib/utils.py", line 238, in fix_unit
    args[name] = float(args[name])
ValueError: could not convert string to float: '100M'
```

A helpful reminder of what units are allowed should be displayed rather than a python traceback, for all arguments that accept units. https://github.com/marcmengel/jobsub_lite/blob/f0bec4f56ed6eb8ff59c0f4c2ffac4e6214b038e/lib/utils.py#L196

PR #157, opened 2022-10-26
draft of fix for #154

marcmengel:
If we fail to convert our value to a float, print a hopefully helpful error message


shreyb (approved):
LGTM

PR #158, opened 2022-11-08
use JobsubCmd rather than Cmd to find the file:/// executable

marcmengel:


shreyb (approved):
Tested by checking out the PR on my dev node.  All looks good.

PR #159, opened 2022-11-08
fix test

marcmengel:
Small fix to test, to make sure we don't have a proxy in X509_USER_PROXY when asked to  make one.


shreyb (approved):
Test looks good.

PR #160, opened 2022-11-18
Verbose/Debug cleanup and Send proxy with jobs

shreyb:
There are two main items in this PR:

1.  Implement the decision that we have a `--verbose <int>` flag, with different levels, and a `--debug` flag, such that `--debug` = `--verbose 1`.  This was done to maintain backward-compatibility with the legacy jobsub_client.
2. Send a 167-hour proxy with jobs by invoking the `x509userproxy` classad in the job definition file.


shreyb:
Draft until my last tests finish running (where I have the job sleep for 6 hours, and check the lifetime of the proxy on either end).

shreyb:
Sleep job worked.   Moving PR to "Ready for Review"

bin/jobsub_submit
```diff
@@ -67,7 +67,7 @@ def render_files(
     """use jinja to render the templates from srcdir into the dest directory
     using values dict for substitutions
     """
-    if values["debug"] > 0:
+    if values["debug"]:
```
marcmengel, line 70: Shouldn't this be checking values["verbose"] now?

marcmengel (approved):
Looks good, one comment about debug vs verbose in jobsub_submit.

bin/jobsub_submit
shreyb: Yes - you're right.  Because if `--debug` is set, that implies `--verbose 1`, but the reverse is not true.  I'll change that.

shreyb: Alright - these are fixed.  Can you please check to make sure it looks good to you now?

marcmengel (approved):
Looks good!

Issue #162, opened 2022-11-22
Extend DAG to allow serial tag inside parallel tag


vitodb:
This is more a feature request than an actual issue.
jobsub_client allowed the use of `serial` tags inside `parallel` tag. This feature was used extensively by the CI.
With jobsub_lite I get the following message:
`Error: file True line 2: <serial> inside <parallel> not currently supported`

marcmengel:
Fixed in #165

Issue #163, opened 2022-11-22
jobsub_lite option `--dag` creates malformed DAG


vitodb:
Using an XML DAG like the following one:
```
<serial>
jobsub -e SLEEP1 -n -G sbnd --expected-lifetime=5m --memory=500MB --disk=1GB --onsite-only -N 1 -l +CI_VALIDATION_STAGE="test_vito_01" file:///usr/bin/sleep ${SLEEP1}
jobsub -e SLEEP2 -n -G sbnd --expected-lifetime=5m --memory=500MB --disk=1GB --onsite-only -N 1 -l +CI_VALIDATION_STAGE="test_vito_02" file:///usr/bin/sleep ${SLEEP2}
jobsub -e SLEEP3 -n -G sbnd --expected-lifetime=5m --memory=500MB --disk=1GB --onsite-only -N 1 -l +CI_VALIDATION_STAGE="test_vito_03" file:///usr/bin/sleep ${SLEEP3}
</serial>
```
creates a DAG with few issues.

1. the `dag.dag` file has lines with `#` that condor doesn't like. Also the line ` PARENT` is appended to the previous line, this makes an invalid DAG. Adding some extra empty line could help to give the DAG better readability.
2. the `stage_N.cmd` files have incorrect lines, the diff output with the correct cmd file has
``` 
23,24c23,24
< transfer_input_files = sleep,simple.cmd
< transfer_output_files = simple.cmd,simple.sh,sleep
---
> transfer_input_files = sleep,stage_1.cmd
> transfer_output_files = stage_1.cmd,stage_1.sh,sleep
```
i.e. it has `simple` instead of `stage_N`

3. similarly `stage_N.sh` files have  `simple` instead of `stage_N`

marcmengel:
Fixed in #165.

PR #165, opened 2022-11-28
implement #162

marcmengel:
Push chunks of dagnabbit parser state on stack; also keep list of start and end of each parallel branch, since serial within parallel has a different end than start job for each line.  Also clean up condor_submit_dag wrapper a bit.


marcmengel:
Also added fixes for #163, mainly small template changes.

marcmengel:
Also including a fix for the discovered bug that we were looking for the pushed proxy in /tmp rather than the current directory in DAGs, which is  unhappy.   We could get rid of that check by copying the proxy into the submit sandbox and just always make it local, but it's a smaller change to make the template smarter...

marcmengel:
Okay, so now we take:
```
jobsub -n --mail_on_error $SUBMIT_FLAGS  ${JOBSUB_EXPORTS} file://./jobA.sh
<parallel>
jobsub -n --mail_on_error $SUBMIT_FLAGS  ${JOBSUB_EXPORTS} file://./jobB.sh
<serial>
jobsub -n --mail_on_error $SUBMIT_FLAGS  ${JOBSUB_EXPORTS} file://./jobC.sh
jobsub -n --mail_on_error $SUBMIT_FLAGS  ${JOBSUB_EXPORTS} file://./jobD.sh
jobsub -n --mail_on_error $SUBMIT_FLAGS  ${JOBSUB_EXPORTS} file://./jobE.sh
</serial>
</parallel>
jobsub -n --mail_on_error $SUBMIT_FLAGS  ${JOBSUB_EXPORTS} file://./jobF.sh
</serial>
```
and turn it into 
```
DOT dag.dot UPDATE
JOB stage_1 stage_1.cmd
VARS stage_1 JOBSUBJOBSECTION="1" nodename="$(JOB)"
JOB stage_2 stage_2.cmd
VARS stage_2 JOBSUBJOBSECTION="2" nodename="$(JOB)"
JOB stage_3 stage_3.cmd
VARS stage_3 JOBSUBJOBSECTION="3" nodename="$(JOB)"
JOB stage_4 stage_4.cmd
VARS stage_4 JOBSUBJOBSECTION="4" nodename="$(JOB)"
PARENT stage_3 CHILD stage_4
JOB stage_5 stage_5.cmd
VARS stage_5 JOBSUBJOBSECTION="5" nodename="$(JOB)"
PARENT stage_4 CHILD stage_5
PARENT stage_1 CHILD stage_2 stage_3
JOB stage_6 stage_6.cmd
VARS stage_6 JOBSUBJOBSECTION="6" nodename="$(JOB)"
PARENT stage_2 stage_5 CHILD stage_6
```
which  runs and makes the pretty picture:

dagTest6

bin/jobsub_submit
```diff
@@ -193,10 +193,11 @@ def main():
     #    shutil.copyfile(token, token_dest)
     #    varg["token"] = token_dest
 
+    varg["is_dag"] = True
```
shreyb, line 196: I understand why you're setting `varg["is_dag"]` this way, but I worry that six months from now, we'll see this line and wonder why we've "defaulted" to setting `varg["is_dag"]` to `True`.  Would it be possible, if slightly less elegant to do something along the lines of:

```python
varg["is_dag"] = False
if args.dag or args.dataset_definition or args.maxConcurrent:
    varg["is_dag"] = True
```
or even 

```python
varg["is_dag"] = True if (args.dag or args.dataset_definition or args.maxConcurrent) else False
```

And then having the `if` block on 197?

I could even see having line 196 be 
```python
varg["is_dag"] = False
```
and then setting `varg["is_dag"]` to `True` for every applicable clause in the `if` statement starting in line 197.  

I know this is extra code compared to what you wrote here, I just worry about the future understandability of this bit.

shreyb (changes requested):
Looks good, except for the one comment in `bin/jobsub_submit`.

bin/jobsub_submit
shreyb: Yeah - it's certainly _nicer_ the way you did it here, but I do have concerns about the clarity.

shreyb (approved):
Looks good - thanks!

marcmengel:
Yes, both.

PR #169, opened 2022-12-01
Dataset dag posts

marcmengel:
This is the fix for the issue Vito mentioned yesterday in the meeting; adding the POST actions in the dataset dags that just succeed.   Stole the items from old jobsub.


marcmengel:
Also some test cleanups (complex DAG test, turning off Dune GP launch tests) and fixing proxy path for dagbegin/dagend.

shreyb (approved):
Good to go.

PR #170, opened 2022-12-01
Make sure voms extension also has 167 hours for generated proxies at submission

shreyb:
Added -vomslife attribute so that the generated proxies have voms extensions that are valid for 167 hours.


marcmengel (approved):
Looks good.

Issue #172, opened 2022-12-05
Incorrect help message in jobsub_q


shreyb:
@rlcee noticed this in his testing:

jobsub_q help says "Only information about jobs owned by the current user will be returned." but that is not true:
dunepro ID: 57106746 10/5 11:04 _ _ 1 _ 1 57106

I confirmed:

```
[sbhat@fifeutilgpvm01 ~]$ jobsub_q --help
usage: jobsub_q [-h] [-G GROUP] [--role ROLE] [--subgroup SUBGROUP]
                [--verbose VERBOSE] [--debug] [-J JOBID] [-name NAME]
                [--jobsub_server JOBSUB_SERVER]

optional arguments:
  -h, --help            show this help message and exit
  -J JOBID, --jobid JOBID
                        job/submission ID
  -name NAME            Set schedd name
  --jobsub_server JOBSUB_SERVER
                        backwards compatability; ignored

general arguments:
  -G GROUP, --group GROUP
                        Group/Experiment/Subgroup for priorities and
                        accounting
  --role ROLE           VOMS Role for priorities and accounting
  --subgroup SUBGROUP   Subgroup for priorities and accounting. See
                        https://cdcvs.fnal.gov/redmine/projects/jobsub/wiki/
                        Jobsub_submit#Groups-Subgroups-Quotas-Priorities for
                        more documentation on using --subgroup to set job
                        quotas and priorities
  --verbose VERBOSE     Turn on more information on internal state of program.
                        --verbose 1 is the same as --debug
  --debug               dump internal state of program (useful for debugging)

also condor_q arguments: [general-opts] [restriction-list] [output-opts | analyze-opts]
(with single '-' or double '--' dashes)

    [general-opts] are
        -global                  Query all Schedulers in this pool
        -schedd-constraint       Query all Schedulers matching this constraint
...a whole bunch of stuff omitted...

    Only information about jobs owned by the current user will be returned.
This default is overridden when the restriction list has usernames and/or
job ids, when the -submitter or -allusers arguments are specified, or
when the current user is a queue superuser
```

The incorrect message is coming from the condor_q help we pass through.  We need to see if we can suppress that bit, since it's wrong, as Ray indicated, and actually against what we want to do.

PR #175, opened 2022-12-05
more jobsub_q tweaks

marcmengel:
Fix for #172 and also making `jobsub_q -G fermilab mengel`  (or other usernames) work right.

### DIFF
--- a/bin/jobsub_cmd
+++ b/bin/jobsub_cmd
@@ -113,7 +113,7 @@ def main() -> None:
                 "-factory",
                 "-unmatchable",
             ]
-            or i[0] in "0123456789"
+            or i[:1].isalnum()
         ):
             default_constraint = False
 
--- a/lib/get_parser.py
+++ b/lib/get_parser.py
@@ -437,4 +437,10 @@ def get_condor_epilog() -> str:
     )
     epilog_l[0] += "(with single '-' or double '--' dashes)\n"
 
+    if condor_cmd == "condor_q":
+        for i in range(len(epilog_l)):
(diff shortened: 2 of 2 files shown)


bin/jobsub_cmd
```diff
@@ -113,7 +113,7 @@ def main() -> None:
                 "-factory",
                 "-unmatchable",
             ]
-            or i[0] in "0123456789"
+            or i[:1].isalnum()
```
shreyb, line 116: Any reason  this needs to be i[:1] rather than i[0]?  Not saying I object, but just wondering.

lib/get_parser.py
```diff
@@ -437,4 +437,10 @@ def get_condor_epilog() -> str:
     )
     epilog_l[0] += "(with single '-' or double '--' dashes)\n"
 
+    if condor_cmd == "condor_q":
```
shreyb, line 440: Clever!  Could you please add a comment here saying why we need to do this?  I'm afraid that six months from now, we'll have no idea why we had to replace part of the condor epilog.

shreyb (changes requested):
Good, apart from my one request in get_parser.py

bin/jobsub_cmd
marcmengel: In my mind f[0] was a character not a string... f[:1] is  a string ... but f[0] 
still is a  string so It doesn't matter...

marcmengel: Which is to say, too much C++ lately :-)

Issue #185, opened 2022-12-06
jobsub_q not showing condor_dagman job


shreyb:
While working on #179, I found an issue with jobsub_q and DAGs.  Namely, the condor_dagman job doesn't show up in jobsub_q:

```
$ jobsub_submit -G fermilab  --dag  file:///home/sbhat/TestDAG/testdag2.dag 

-----------------------------------------------------------------------
File for submitting this DAG to HTCondor           : /home/sbhat/.cache/jobsub_lite/js_2022_12_06_231511_1309ef34-f8fa-44cf-8897-4a74de3598cb/testdag2.dag.condor.sub
Log of DAGMan debugging messages                 : /home/sbhat/.cache/jobsub_lite/js_2022_12_06_231511_1309ef34-f8fa-44cf-8897-4a74de3598cb/testdag2.dag.dagman.out
Log of HTCondor library output                     : /home/sbhat/.cache/jobsub_lite/js_2022_12_06_231511_1309ef34-f8fa-44cf-8897-4a74de3598cb/testdag2.dag.lib.out
Log of HTCondor library error messages             : /home/sbhat/.cache/jobsub_lite/js_2022_12_06_231511_1309ef34-f8fa-44cf-8897-4a74de3598cb/testdag2.dag.lib.err
Log of the life of condor_dagman itself          : /home/sbhat/.cache/jobsub_lite/js_2022_12_06_231511_1309ef34-f8fa-44cf-8897-4a74de3598cb/testdag2.dag.dagman.log

-no_submit given, not submitting DAG to HTCondor.  You can do this with:
"condor_submit /home/sbhat/.cache/jobsub_lite/js_2022_12_06_231511_1309ef34-f8fa-44cf-8897-4a74de3598cb/testdag2.dag.condor.sub"
-----------------------------------------------------------------------
Submitting job(s).
1 job(s) submitted to cluster 57107096.
Use job id 57107096.0@jobsub01.fnal.gov to retrieve output
$ ./jobsub_lite/bin/jobsub_q -G fermilab 57107096.0@jobsub01.fnal.gov
JOBSUBJOBID                             OWNER           SUBMITTED     RUNTIME   ST PRIO   SIZE  COMMAND
57107096.0@jobsub01.fnal.gov            sbhat           12/06 23:15   0+00:00:01 I    0    0.3 condor_dagman -p 0 -f -l . -Lockfi
[...]

Issue #179, opened 2022-12-06
X509 Proxy gets delegated by schedd


shreyb:
Even though we've set the schedd to not delegate proxies, they show up as 24-hour delegated proxies in jobs, even though they're originally week-long.

Nick Peregonow and I figured out that we need to add the following line to each JDF template where `+x509userproxy` appears:

```delegate_job_GSI_credentials_lifetime = 0```

Since submitting remotely, as we do, automatically tells the schedd to delegate the proxy (due to the file transfer code being run, as https://lists.cs.wisc.edu/archive/htcondor-users/2013-September/msg00077.shtml says) we need to set this on each job at submit time so that the proxy gets the full lifetime.

See https://htcondor.readthedocs.io/en/v8_8/admin-manual/configuration-macros.html#index-1006 for more details.

shreyb:
Test parameters:

1.  Make sure that all unit tests pass
2. Submit a job in fermilab group.  Make sure the following:
   1.  `voms-proxy-info -all -file /tmp/x509up_u$(id -u)-fermilab_Analysis | grep timeleft` gives two lines with roughly 167:59:00 left
   2. `condor_q -G fermilab -name <schedd> <jobid> -af x509userproxyexpiration` gives a time for next week
3. Repeat (2) with a DAG job

shreyb:
1,2 work.  Stuck on (3).  Will talk to Marc about this.

shreyb:
Going to split this into two parts - part 1 will be regular submits, part 2 DAGs.  This is because part 2 relies on #185 getting fixed.

shreyb:
Forgot to note:  part 1 works.

shreyb:
part-2 in progress now.  (3) from the testing procedure now works.  Just waiting for DAG to run to submit this as PR.

PR #186, opened 2022-12-07
Add line in job JDF to tell condor schedd not to curtail lifetime of x509 proxy

shreyb:
Add line in job JDF to tell condor schedd not to curtail lifetime of x509 proxy, along with a couple of verbose statements and small changes to increase readability.

Passed unit tests, and test laid out in #179.

This solves the first half of #179, but does not fix the issue with DAGs.  When #185 is completed, I will implement the same change to the DAG templates in a future PR.

### DIFF
--- a/templates/simple/simple.cmd
+++ b/templates/simple/simple.cmd
@@ -87,6 +87,7 @@ use_oauth_services = {{group}}
 {% else %}
 +x509userproxy = "{{proxy}}"
 {% endif %}
+delegate_job_GSI_credentials_lifetime = 0
 {% endif %}
 
 queue {{N}}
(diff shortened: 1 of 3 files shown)


templates/simple/simple.cmd
```diff
@@ -87,6 +87,7 @@ use_oauth_services = {{group}}
 {% else %}
 +x509userproxy = "{{proxy}}"
 {% endif %}
+delegate_job_GSI_credentials_lifetime = 0
```
marcmengel, line 90: Do we also need this in the various templates/dag*/*dag.sub files?

marcmengel (approved):
Approved modulo the comment below...

templates/simple/simple.cmd
shreyb: Yes - but that will happen after #185 is done, if you're OK with that.

PR #190, opened 2022-12-08
Add --onsite and --offsite-only flags

shreyb:
Marc - could you please make sure I didn't mess anything up in the test_get_parser.py file?  I think I understood how it figures out the arguments, but a second pair of eyes on that bit especially would be very heflpul.


marcmengel (approved):
That looks like a good way to handle the aliased arguments.

PR #191, opened 2022-12-08
Change key access to get() calls where we may not have a var set from a caller

shreyb:
Avoid bugs like the one that affected 0.2 by using get() calls instead of straight key-accesses.


marcmengel (approved):
LGTM

Issue #182, opened 2022-12-06
jobsub_q should be able to filter by Owner of job


shreyb:
In jobsub_client, we have an option "--user" that can filter the results of jobsub_q  by user.  We should have that in jobsub_lite.

marcmengel:
condor_q and current jobsub_q take just the username, so we could add a no-op --user flag...

shreyb:
Ooh - that would be nice.  Though we should be careful, because you can ALSO say `condor_q <jobid>`.  I think a no-op flag would still be OK.  If we wanted to be super-safe, we could of course have the --user flag create a constraint for the Owner.

PR #192, opened 2022-12-09
adding --user username flag do jobsub_q

marcmengel:
Implements #182


shreyb (changes requested):
It looks like you forgot to do a `git submodule update` - the config submodule moved backwards a commit compared to the current master (it should be b18329f, and it's 5ce0a7f in this PR).

marcmengel:
Did a git pull in the submodule... git update actually kicks it back to what it thought the revision was...

shreyb (approved):

PR #193, opened 2022-12-12
Don't curtail X509 proxy lifetime in DAGs

### LINKED ISSUE #179: X509 Proxy gets delegated by schedd

shreyb:
This closes #179, implementing the JDF classad attribute needed to not curtail X509 proxy lifetime in DAGs.  We already did this for regular jobs in #186.

@marcmengel - could you please test this on a dataset DAG?  I did already for a regular DAG, but I'd like to make sure this doesn't break more complex stuff.  Thanks!

### DIFF
--- a/templates/dag/dag.dag.condor.sub
+++ b/templates/dag/dag.dag.condor.sub
@@ -26,6 +26,7 @@ environment	=  _CONDOR_SCHEDD_ADDRESS_FILE=/var/lib/condor/spool/.schedd_address
 +JobsubClientVersion="{{jobsub_version}}"
 +JobsubClientKerberosPrincipal="{{kerberos_principal}}"
 x509userproxy = {{proxy}}
+delegate_job_GSI_credentials_lifetime = 0
 {{lines|join("\n")}}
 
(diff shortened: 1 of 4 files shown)


marcmengel (commented):
I have your branch checked out... tests are running..

marcmengel (approved):
Tests successful.  Looks good to me.

shreyb:
Thanks Marc!

PR #194, opened 2022-12-13
JOBSUB_TEST_INSTALLED environment flag

marcmengel:
Added support for doing:

JOBSUB_TEST_INSTALLED=1 pytest

to test the installed version of jobsub, rather than the copy we're sitting in.

### DIFF
--- a/tests/test_condor_unit.py
+++ b/tests/test_condor_unit.py
@@ -12,8 +12,12 @@ os.chdir(os.path.dirname(__file__))
 
 #
 # import modules we need to test, since we chdir()ed, can use relative path
+# unless we're testing installed, then use /opt/jobsub_lite/...
 #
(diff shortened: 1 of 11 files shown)


marcmengel:
Trying to test this by putting a copy of the test subdirectory completely away from the rest of the repo and
running tests...  Not happy so far :-(  May need another patch...

marcmengel:
Okay, so now with that latest commit all the tests look good if you have just the "jobsub_lite/tests" subtree, and you set
JOBSUB_TEST_INSTALLED=1 in the environment.

shreyb (approved):
Looks good.  So the idea here is that people can download the repo onto a machine that has jobsub_lite, set the environment variable, and then run the tests with the installed version?

Issue #196, opened 2022-12-19
dropbox bug


vitodb:
Copied from Herb Greenlee ticket INC000001144465 in ServiceNow:

```
Outage start date: 2022-12-19 09:06:08 (US/Central)
Outage/Incident details:
This is a bug report about jobsub_lite 0.3.  Option -f dropbox:// does not work with relative paths.

fifeutilgpvm02:~ $ ls -l test*.dat
-rw-r--r-- 1 greenlee greenlee 2 Nov 21 16:12 test3.dat
-rw-r--r-- 1 greenlee greenlee 2 Nov 21 16:12 test4.dat
fifeutilgpvm02:~ $ jobsub_submit -G uboone -f dropbox:///home/greenlee/test3.dat -f dropbox://test4.dat file:///home/greenlee/bin/testbatch.sh
test3.dat
Using bearer token located at /tmp/bt_token_uboone_Analysis_1731 to authenticate to RCDS
tar: test4.dat: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
Using bearer token located at /tmp/bt_token_uboone_Analysis_1731 to authenticate to RCDS
Submitting job(s).
1 job(s) submitted to cluster 57107229.
Use job id 57107229.0@jobsub01.fnal.gov to retrieve output

I can't think of a reason why "-f dropbox://" shouldn't work with relative paths (as opposed to non-dropbox option "-f").
```
To note that also the syntax `-f dropbox://./test4.dat` has the same issue.

marcmengel:
Note that dropbox:file.dat works.  But we should definitely handle the slashes
like old jobsub.

shreyb:
@marcmengel  - is this ready to close?

marcmengel:
Yes, I think so.

shreyb:
Great!  @vitodb we'll go ahead and close this.  If you'd like to check out this version of the code and try it out, feel free (I can help with that if you need), but otherwise, we're planning on another release right after the break and before go-live.

PR #197, opened 2022-12-19
tarfile hash repeatability fix

marcmengel:
[from slack channel]
vito  11:48 AM
while doing some check for jobsub_lite ticket INC000001144465 [dropbox bug] it looks like if I want to transfer a file to the job I get different hash every time I use that file, so RCDS picks up a different CVMFS path each time, is this the expected behavior?

mengel [12:09 PM]
It's supposed to be based on the md5 hash of the tarfile... but that tickles a memory: https://unix.stackexchange.com/questions/438329/tar-produces-different-files-each-time
let me see if their workaround works here...
(i.e using gzip -n when compressing the repository to omit the timestamp)

Okay that works... thus this pull request, with a test!

### DIFF
--- a/lib/fake_ifdh.py
+++ b/lib/fake_ifdh.py
@@ -215,8 +215,17 @@ def getProxy(
     return vomsfile
 
 
+def fix_pnfs(path: str) -> str:
+    m = re.match(r"/pnfs/(.*)", path)
+    if m:
+        path = f"https://fndca1.fnal.gov:2880/{m.group(1)}"
+    return path
+
+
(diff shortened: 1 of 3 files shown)


marcmengel:
Hang on, trying to drop a fix for #196 on this pull request as well...

marcmengel:
... and found a pnfs dropbox bug and rolled that in, too... sigh.

shreyb (changes requested):
Happy overall, but I have a few nitpicky comments with future readability in mind.

PR #199, opened 2022-12-23
Fix --site flag

### LINKED ISSUE #198: Don't add ", Fermigrid" if OFFSITE is specified/Safeguard Sites input by stripping trailing commas/spaces in --site argument
In this clause:
https://github.com/marcmengel/jobsub_lite/blob/3b666fadb6446f1bdde01f988d4da3b999ed32a0/lib/utils.py#L147

We simply pass the list onto Condor.  That's fine, but we need to guard against cases like this where the user has an extra comma or space, by just stripping those out.

```
Slack conversation:
kherner:
And the --site option no longer works, right?
I'm doing new site testing so I figured I would kill two birds with one stone on  jobsub_lite tests.

Shreyas
no - --site  should work

kherner
It didn't for 57107238.0@jobsub01.fnal.gov
+DESIRED_SITES = "CA_SFU, "
Any chance the comma, space at the end messes up the parsing?

Shreyas
I think I found it
we don't actually parse that out
we just appended it as is
"DESIRED_SITES = "CA_SFU, " is in your job's classad
so that's a bug
```

shreyb:
There's a lot here.  This is a summary of the changes.

1.  --site now works
2. --site defaults to "", not None
3. --site is mutually exclusive with --onsite/--offsite flags
4. As described in #198, there is new logic on how --site, --onsite/offsite, and --resource-provides=usage_model is handled.  The summary is here:  https://github.com/marcmengel/jobsub_lite/issues/198#issuecomment-1363266689 .  We basically give precedence to --site, then --onsite/offsite, then --resource-provides=usage_model.
5. Changed templates to select site properly

I've requested all three of you to check this:
@goodenou: this reimplements the logic you had before, but slightly changed.  Please make sure it makes sense to you, both from a technical and as FIFE lead in terms of policy.
@retzkek: (3) above - does it look right to you?
@marcmengel:  Testing.  I've done a lot of testing of this PR, (see #198), but would you be able to check it with DAGs?  I've done some basic DAG testing, but it's not really my forte.

Of course, as of this writing, everyone's on vacation, so please feel free to review this when you're back.

Thanks all!

Closes #198

### DIFF
--- a/lib/get_parser.py
+++ b/lib/get_parser.py
@@ -331,9 +331,6 @@ def get_parser() -> argparse.ArgumentParser:
         " '&&(CVMFS==\"OSG\")' to the job requirements",
     )
     parser.add_argument(
-        "--site", default="", help="submit jobs to these (comma-separated) sites"
-    )
-    parser.add_argument(
         "--tar_file_name",
         "--tar-file-name",
         default=[],
@@ -390,6 +387,12 @@ def get_parser() -> argparse.ArgumentParser:
     )
 
     usage_model_group = parser.add_mutually_exclusive_group()
+    usage_model_group.add_argument(
+        "--site",
+        type=str,
+        default="",
+        help="submit jobs to these (comma-separated) sites",
+    )
     usage_model_group.add_argument(
         "--onsite",
         "--onsite-only",
--- a/lib/get_parser.py
+++ b/lib/get_parser.py
@@ -330,7 +330,9 @@ def get_parser() -> argparse.ArgumentParser:
         ' +DESIRED_CVMFS="OSG" to the job classad attributes and'
         " '&&(CVMFS==\"OSG\")' to the job requirements",
     )
-    parser.add_argument("--site", help="submit jobs to these (comma-separated) sites")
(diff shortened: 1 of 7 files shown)


marcmengel:
So, testing this change with a --dataset-definition DAG:
Before change, jobs run with --site=Wisconsin have:
```
DESIRED_SITES = "Fermigrid, "
DESIRED_usage_model = "OPPORTUNISTIC,DEDICATED,OFFSITE"
```
and run on-site...

After the change they have:
```
DESIRED_SITES = "Wisconsin"
DESIRED_usage_model = "OFFSITE"

```
waiting to see where it matches...  

Okay, so on jobsub01 this works fine.  dagman-launched jobs from jobsubdevgpvm01 seem to be having trouble starting at Wisconsin (firewall?!?), but I don't think that's related.

marcmengel (approved):
Testing completed.  Works well on jobsub01.  
Dag jobs seem to have trouble launching on jobsubdevgpvm01, but I think that's some kind of firewall problem(?) when the remote condor_startd tries to call back into the schedd...

PR #202, opened 2023-01-05
Weakened tokens

marcmengel:
This adds code to add weakened token scopes to the submissions with command line flags;


lib/token_mods.py
```diff
@@ -0,0 +1,83 @@
+import os
+import os.path
+import shutil
+from typing import List, Set
+
+
+def get_job_scopes(
+    tokenfile: str, need_modify: List[str], need_scopes: List[str]
+) -> List[str]:
```
shreyb, line 9: Just like the other functions in this module, could you please add a quick comment/docstring saying what this function does?  From what I gather, it's to add modify scopes to a token, if they're allowed, or any other scope that's needed.  Is that right?

lib/get_parser.py
```diff
@@ -234,6 +234,18 @@ def get_parser() -> argparse.ArgumentParser:
         default=[""],
         help="Lines to append to the job file.",
     )
+    parser.add_argument(
+        "--need-modify",
```
shreyb, line 238: I realized in my writeup, I called this `--need-modify`, but should we actually make that `--need-storage-modify` since there are modify scopes for other types (like `compute.modify`)?  What do you think?

bin/jobsub_submit
```diff
@@ -171,19 +173,37 @@ def main():
...
+    # different one for different scopes
+    #
+    token = use_token_copy(token)
+    varg["job_scope"] = " ".join(
+        get_job_scopes(token, args.need_modify, args.need_scope)
+    )
+    m = hashlib.sha256()
+    m.update(varg["job_scope"].encode())
+    varg["oauth_handle"] = m.hexdigest()[:10]
+
```
shreyb, line 206: Nice.

shreyb (changes requested):
Just a couple of small things I had questions/comments about.  Otherwise, all good.

lib/get_parser.py
marcmengel: Making this change...

marcmengel: This is in now.

shreyb:
Btw, @marcmengel - is this documented anywhere besides the issue?  If not, we should probably put something up in the Getting Started wiki.

marcmengel:
I put a note in the Differences between jobsub and jobsub_lite page, but I'll copy the section into the Getting Started page as well.

marcmengel:
I tacked it on the bottom under "If you have problems"...

PR #204, opened 2023-01-09
Project name

marcmengel:
--project_name flag

### DIFF
--- a/lib/get_parser.py
+++ b/lib/get_parser.py
@@ -227,7 +227,10 @@ def get_parser() -> argparse.ArgumentParser:
         " jobs in a DAG",
     )
     parser.add_argument(
-        "--jobsub_server", default="", help="Backwards compatability flag. Ignored"
+        "--jobsub-server",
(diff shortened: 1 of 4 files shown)


lib/get_parser.py
```diff
@@ -226,6 +226,12 @@ def get_parser() -> argparse.ArgumentParser:
         help="generate and mail a summary report of completed/failed/removed"
         " jobs in a DAG",
     )
+    parser.add_argument(
```
shreyb, line 229: We're discussing this offline.

tests/test_get_parser_unit.py
```diff
@@ -150,6 +150,8 @@ def all_test_args():
         "--group",
         "xxgroupxx",
         "--jobid",
+        "--jobsub-server",
```
shreyb, line 153: Same with this.

shreyb (commented):
Discussing the `--jobsub-server` flag offline.

PR #205, opened 2023-01-09
add version flag to jobsub_

goodenou:
Added version.py file with some general information in it, including the version of the code. Command to get the version is `Jobsub_ --version` or `Jobsub_ -v`. Output has the format "jobsub_lite version 0.3"

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -109,6 +109,10 @@ def set_extras_n_fix_units(
     args["kerberos_principal"] = get_principal()
     args["uid"] = str(os.getuid())
 
+    for i in range(len(args["lines"])):
+        # do 1 layer of \x -> x to be compatible with jobsub_client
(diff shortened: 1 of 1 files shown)


shreyb:
pre-commit.ci autofix

bin/jobsub_
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/python3 -I
```
shreyb, line 1: Rather than have a separate executable to get the version, would it be possible to just add these flags (lines 46-57 here) into the lib/get_parser.py file so all jobsub_* executables inherit it?  

If so, I think it should be in the base parser:
https://github.com/marcmengel/jobsub_lite/blob/master/lib/get_parser.py#L102

What do you think?

shreyb (changes requested):

PR #209, opened 2023-01-11
sigh... turning weakened tokens back off 

marcmengel:
until we figure out authentication issue for production/POMS submisssion (kcron?)


shreyb (changes requested):
This looks good, but would it be possible to comment out the code in jobsub_submit that looks at token_mods.py, and/or put a note in there that it currently doesn't do anything, but is there for a future implemented feature?

marcmengel:
Well if we comment out the calls, we won't have token_scope defined, and we'll have to actually remove the lines from the templates rather than comment them out...

shreyb:
Got it. Then it's fine as is. Feel free to merge this - I'm away from a computer for the next few hours.

Get Outlook for Android<https://aka.ms/AAb9ysg>

shreyb (approved):

PR #212, opened 2023-01-12
do backslash-escape processing on all args, not just --lines

marcmengel:
This is round 2 of addressing the compatability issue that Heidi brought up.  
In the old jobsub_client, the jobsub_submit arguments were sent over to the server, and
then the server ran a server-side jobsub_submit script locally, which involved running
another shell, and an extra eval-layer of substituting backslash-whatever. 
This happened to all of the command-line arguments, so here we do the whole of sys.argv.


shreyb (approved):
LGTM

Issue #17, opened 2022-03-17
Add --version flag to jobsub commands


shreyb:
jobsub_<cmd> --version should give the version of jobsub_lite.

shreyb:
@rlcee has brought up that we should make sure that the output string of this flag should have "jobsub_lite" in there, to make it clear to users they're using a jobsub_lite version, not old jobsub_client.

shreyb:
Closed by #213

PR #213, opened 2023-01-12
adding version and help email to jobsub_ arguments

goodenou:
I added arguments "--version" and ""support-email" to the list of base arguments so that folks can now submit a command such as "jobsub_fetchlog --version" and get the following output back "jobsub_lite version 0.3". The version is stored in lib/version.oy.

### DIFF
--- a/bin/jobsub_submit
+++ b/bin/jobsub_submit
@@ -51,6 +51,7 @@ from tarfiles import do_tarballs
 from utils import set_extras_n_fix_units, cleanup
 from creds import get_creds
 from token_mods import get_job_scopes, use_token_copy
+import version
 
 
 def get_basefiles(dlist: List[str]) -> List[str]:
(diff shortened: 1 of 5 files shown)


bin/jobsub_submit
```diff
@@ -170,6 +171,14 @@ def main():
     parser = get_parser()
     args = parser.parse_args()
 
+    if args.version:
```
shreyb, line 174: I like this.  Depending on if we want to expand this to other commands, we might move this code into get_parser as a callback class/function, but this works well for what we expect folks to do.

shreyb (commented):
Leaving it unapproved while I take a look at the CI failures.  But the code looks good.

shreyb:
pre-commit.ci autofix

shreyb:
Closes #17

PR #214, opened 2023-01-12
Jobsub submit dag

shreyb:
Per Herb Greenlee's request, this is a jobsub_submit_dag executable.

There were also some broken tests from previous PRs, so I fixed those here.


marcmengel (approved):
Looks good to me.

PR #215, opened 2023-01-12
Add history

marcmengel:
Add a jobsub_history command, implemented with condor_history.

Did in the style of jobsub_q, so you can also use all the condor_history flags.

### DIFF
--- a/bin/jobsub_history
+++ b/bin/jobsub_history
@@ -38,7 +38,6 @@ import get_parser
 def main() -> None:
     """main line of code, proces args, etc."""
     parser = get_parser.get_jobid_parser(add_condor_epilog=True)
-    parser.add_argument("-name", help="Set schedd name", default="")
     parser.add_argument("--user", help="Set username to look at", default="")
     parser.add_argument(
         "--qdate-ge",
@@ -54,10 +53,6 @@ def main() -> None:
     arglist, passthru = parser.parse_known_args()
 
     passthru.append("-backwards")
-
-    if arglist.name:
-        passthru.append("-name")
-        passthru.append(arglist.name)
 
     # Re-insert --debug/--verbose if it was given
     if arglist.verbose:
@@ -117,6 +112,7 @@ def main() -> None:
     if schedd:
         passthru.append("-name")
         passthru.append(schedd)
+
     passthru = out
 
     if default_formatting:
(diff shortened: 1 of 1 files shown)


bin/jobsub_history
```diff
@@ -0,0 +1,149 @@
...
+sys.path.append(os.path.join(PREFIX, "lib"))
+
+import fake_ifdh
+import get_parser
+
+
+def main() -> None:
+    """main line of code, proces args, etc."""
+    parser = get_parser.get_jobid_parser(add_condor_epilog=True)
+    parser.add_argument("-name", help="Set schedd name", default="")
```
shreyb, line 41: Should this be `--name`?

bin/jobsub_history
```diff
@@ -0,0 +1,149 @@
...
+    passthru.append("-backwards")
+
+    if arglist.name:
+        passthru.append("-name")
+        passthru.append(arglist.name)
+
+    # Re-insert --debug/--verbose if it was given
+    if arglist.verbose:
+        passthru.append("-debug")
+    # if they gave us --jobid or --user put in the value plain, condor figures it out
```
shreyb, line 65: But in jobsub_cmd, we have this:
https://github.com/marcmengel/jobsub_lite/blob/007f4f2ac61b650ba3217f91013aaa5d12bfddc2/bin/jobsub_cmd#L102

That specifically parses out things like 1234@jobsubschedd.domain.  Don't we need that kind of thing here around line 90?

shreyb (changes requested):
Just a couple of comments here for your review.

bin/jobsub_history
marcmengel: Actually --name will get turned into -name later... I forget now what the case was that made us add that to the jobsub_q code, but I'm pretty sure we had some reason for it.

bin/jobsub_history
marcmengel: Yep, put it back in.

bin/jobsub_history
marcmengel: Okay, there it is... https://github.com/marcmengel/jobsub_lite/commit/2701f479262a2f7184aa9c83ca6e2fc6f50465cd
so we were doing that so we detect the schedd even if they specify it as -name, because we're setting _condor_CREDD_HOST=schedd in the environment... So we probably do *not* need that in the jobsub_history code.

PR #216, opened 2023-01-13
Add support for pre/post scripts in DAG

vitodb:
Add support for pre/post scripts in DAG using dagnabbit.
This also include a dagTest8 template and a pre and post sample scripts.


lib/dagnabbit.py
```diff
@@ -183,6 +188,64 @@ def parse_dagnabbit(
...
+                    sys.stderr.write(f"Error at file {dagfile} line {linenum}\n")
+                    sys.stderr.write(f"parsing: {line.strip().split()}\n")
+                    sys.stderr.flush()
+                    raise
+                prescript = line.split()[1].replace("file://", "")
+                prescript_base = os.path.basename(prescript)
+                prescript_args = " ".join(line.split()[2:])
+                of.write(f"SCRIPT PRE {name} {prescript_base} {prescript_args}\n")
+                thesevalues["prescript"] = prescript
+                if "postscript" in thesevalues:
```
marcmengel, line 215: This could use a comment -- why do we delete the postscript if we see a prescript (and vice versa, below)?

marcmengel (changes requested):

lib/dagnabbit.py
vitodb: Sure, I can add a comment in the code here.
Possibly there is a better way to get the same result.
Here, when processing postscript, we need to delete prescript otherwise in utils.py:set_extras_n_fix_units we end up to copy the prescript script to itself at its destination raising an error.
Maybe it could make the code easier to follow by deleting `thesevalues["prescript"]` after the `set_extras_n_fix_units` call (and similarly for postscript)?

vitodb: Possibly commit [5887b13](https://github.com/marcmengel/jobsub_lite/pull/216/commits/5887b1365a583be7609bc73569599df67e89c6aa) has a better implementation of handling pre/post script rather than deleting its reference?

marcmengel (approved):
I like it.

Issue #223, opened 2023-01-17
Fix documentation for cvmfs tarball variable, provide symlink 


shreyb:
Our documentation for the `--tar-file-name` argument has been wrong for quite some time (including in the old jobsub_client).  It states that `--tar-file-name` will send files to `$INPUT_TAR_FILE` in the job.  

https://github.com/marcmengel/jobsub_lite/blob/fc5e6a18c78c9a3bfde1054c868db766cb9a181d/lib/get_parser.py#L370

This is wrong for the RCDS (now default) case, since RCDS automatically unpacks the tarfile, and we point the environment variable `$INPUT_TAR_DIR_LOCAL` to that unpacked directory on cvmfs.

https://github.com/marcmengel/jobsub_lite/blob/fc5e6a18c78c9a3bfde1054c868db766cb9a181d/templates/simple/simple.sh#L217

So we need to do four things here:

* change documentation to mention that if using RCDS (which is default), look for unpacked tarball in `$INPUT_TAR_DIR_LOCAL`
* symlink in `$CONDOR_DIR_INPUT` to point to either `$INPUT_TAR_DIR_LOCAL` or `$INPUT_TAR_FILE` to provide "backward compatibility" for old behavior, where tarball was unwound into `$CONDOR_DIR_INPUT/z` for a tarball `z.tar`
* Update this wiki page to reflect this:  https://fifewiki.fnal.gov/wiki/Differences_between_jobsub_lite_and_legacy_jobsub_client/server
*  Update https://github.com/marcmengel/jobsub_lite/wiki#jobsub_submit with the updated `jobsub_submit` usage.

shreyb:
@marcmengel and I came up with a summary of the intended behavior:

* -f:  send file to `$CONDOR_DIR_INPUT`, don't unpack (ALREADY DONE)
* -f dropbox or tardir:  use RCDS or PNFS, tar up any file, even if it's already a tarball for dropbox, send to `$CONDOR_DIR_INPUT`, and don't unpack (ALREADY DONE)
* --tar-file-name anything:  use RCDS or PNFS: 
  * for RCDS, set `$INPUT_TAR_DIR_LOCAL` and `$INPUT_TAR_FILE` to RCDS dir where this exists (since there is no tarball); or 
  * for PNFS:  `ifdh cp`PNFS tarball in, unpack, set $INPUT_TAR_FILE to tarball, $INPUT_TAR_DIR_LOCAL to unpacked tarball directory.  
[...]

PR #224, opened 2023-01-18
fixes for compatability with old jobsub regarding tarfiles

marcmengel:
… and dropbox:, also fix for -f dropbox://file   This allso covers the symlink half of #223

### DIFF
--- a/templates/simple/simple.sh
+++ b/templates/simple/simple.sh
@@ -219,6 +219,9 @@ ${JSB_TMP}/ifdh.sh cp -D {{fname}} ${CONDOR_DIR_INPUT}
     {%if loop.first%}
       INPUT_TAR_DIR_LOCAL={{tfname}}
       export INPUT_TAR_DIR_LOCAL
+      INPUT_TAR_FILE={{tfname}}
+      export INPUT_TAR_FILE
+      ln -s {{tfname}} ${CONDOR_DIR_INPUT}/{{tar_file_orig_basenames[loop.index0]}}
 
       # wait for tarfile to show up
       num_tries=0
@@ -235,6 +238,9 @@ ${JSB_TMP}/ifdh.sh cp -D {{fname}} ${CONDOR_DIR_INPUT}
     {%else%}
       INPUT_TAR_DIR_LOCAL{{loop.index0}}={{tfname}}
       export INPUT_TAR_DIR_LOCAL{{loop.index0}}
+      INPUT_TAR_FILE_{{loop.index0}}={{tfname}}
+      export INPUT_TAR_FILE_{{loop.index0}}
+      ln -s {{tfname}} ${CONDOR_DIR_INPUT}/{{tar_file_orig_basenames[loop.index0]}}
     {%endif%}
   {%else%}
     # tarfile to transfer and unpack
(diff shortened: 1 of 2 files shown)


templates/simple/simple.sh
```diff
@@ -243,11 +249,17 @@ ${JSB_TMP}/ifdh.sh cp -D {{fname}} ${CONDOR_DIR_INPUT}
     ${JSB_TMP}/ifdh.sh cp {{tfname}} {{tflocal}}
     tar --directory .unwind_{{loop.index0}} -xzvf {{tflocal}}
     {%if loop.first%}
+      INPUT_TAR_DIR_LOCAL=`pwd`/.unwind_{{loop.index0}}
+      export INPUT_TAR_DIR_LOCAL
       INPUT_TAR_FILE=`pwd`/.unwind_{{loop.index0}}
```
shreyb, line 254: Looking over this again, isn't this line inconsistent with the docs (I know you didn't change it now, but I just noticed this)?  

Both #223 and the helptext say that `$INPUT_TAR_FILE` points to the actual tarfile in the PNFS case, not the directory.  So for the CVMFS case, both `$INPUT_TAR_FILE` and `$INPUT_TAR_DIR_LOCAL` point to the directory, since there is no tarfile to point at, but for the PNFS case, `$INPUT_TAR_FILE` points at the actual tarball (so `{{tflocal}}`, it appears), and `$INPUT_TAR_DIR_LOCAL` points to the directory containing the unpacked tarball.

Am I understanding that correctly?  If so, do you think it's a reasonable way to do it, or do you think the PNFS and CVMFS cases should not differ?  Basically, it's a tossup between honoring the meaning of a variable named `$INPUT_TAR_FILE`, and having that point to an actual tarfile if possible, or keeping the behavior consistent no matter what the dropbox method. 

What are your thoughts?

templates/simple/simple.sh
```diff
@@ -243,11 +249,17 @@ ${JSB_TMP}/ifdh.sh cp -D {{fname}} ${CONDOR_DIR_INPUT}
...
       export INPUT_TAR_FILE
+      ln -s $INPUT_TAR_FILE ${CONDOR_DIR_INPUT}/{{tar_file_orig_basenames[loop.index0]}}
     {%else%}
+      INPUT_TAR_DIR_LOCAL{{loop.index0}}=`pwd`/.unwind_{{loop.index0}}
+      export INPUT_TAR_DIR_LOCAL{{loop.index0}}
       INPUT_TAR_FILE_{{loop.index0}}=`pwd`/.unwind_{{loop.index0}}
```
shreyb, line 260: Same comment as above.

shreyb (changes requested):
Looks great, with the exception of the one comment I have in there regarding what `$INPUT_TAR_FILE` should be in the PNFS case.  Based on what we decide, I can do the documentation side if you'd like.

shreyb (approved):
Looks good!  Did your test jobs work with these changes?

marcmengel:
Now the test jobs are working (had to fix the pnfs dropbox directory for 'fermilab' jobs  to be /pnfs/fermilab/users/$USER/
since there is no fermilab persistent area, and the tokens for fermilab only let you see /pnfs/fermilab/users/$USER)

PR #225, opened 2023-01-18
Updated --tar-file-name documentation with corrected behavior

shreyb:
Documentation for change in #223, #224.


marcmengel (changes requested):
At the moment, the later --tar_file_name options will set
INPUT_TAR_FILE_{1,2,3...}
but
INPUT_TAR_DIR_LOCAL{1.2.3}
(no underscore).   We should either add the underscores in the simple.sh template,
 or take them out here. (probably the former, look for "INPUT_TAR_DIR_LOCAL{{loop.index0}}" and add the underscore.).

shreyb:
Yeah - I'd opt for adding the underscore as well.  I'll put that in.

shreyb:
Changed.  Does this look good now, @marcmengel?

marcmengel (approved):

PR #227, opened 2023-01-19
code to read a default role

marcmengel:
Allow for a default role file to be read in that either the managed token service could push
or the user could add.   It is either:
`/tmp/jobsub_default_role_${GROUP}_$(id -u)`
or
`$HOME/.jobsub_default_role_${GROUP}_$(id -u)` 
and should contain the role (i.e. "Analysis" or "Production")

### DIFF
--- a/lib/fake_ifdh.py
+++ b/lib/fake_ifdh.py
@@ -21,6 +21,7 @@
 
 import argparse
 import json
+import glob
 import os
 import re
 import shlex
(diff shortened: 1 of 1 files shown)


lib/fake_ifdh.py
```diff
@@ -57,6 +57,19 @@ def getRole(role_override: Optional[str] = None, verbose: int = 0) -> str:
     if role_override:
         return role_override
 
+    # if we have a default role pushed with a vault token, or $HOME/.jobsub_default... use that
+    uid = os.getuid()
+    group = os.environ["GROUP"]
+
+    for prefix in ["/tmp/", f"{os.environ['HOME']}/."]:
```
shreyb, line 64: I was looking at this, and for the sake of consistency, I think the filename should be the same, whichever directory it's in.  So either `.jobsub_default_role_{group}_{uid}` or `jobsub_default_role_{group}_{uid}`.  Otherwise, people might get confused.  What do you think?

shreyb (approved):
Looks good, aside from the one comment.

lib/fake_ifdh.py
marcmengel: That's fine.  or should we make it $HOME/.config/jobsub_default_role_{group}_{uid} ?  Then it's not clogging up the home directory, and its spelled the same?

shreyb: That's a good idea.

shreyb (approved):
LGTM

Issue #228, opened 2023-01-20
If there's an error with authentication, we should not display "Submitting job(s)"


shreyb:
Like what we see below:

```
mu2egpvm02 02:24 PM ~ > jobsub_submit --role MARS -G accel file:///usr/bin/printenv
Attempting OIDC authentication with https://htvaultprod.fnal.gov:8200/

Complete the authentication at:
    https://cilogon.org/device/?user_code=W22-HR7-L3R
No web open command defined, please copy/paste the above to any web browser
Waiting for response in web browser
Storing vault token in /tmp/vt_u4395
Saving credkey to /nashome/b/boyd/.config/htgettoken/credkey-accel-mars
Saving refresh token ... done
Attempting to get token from https://htvaultprod.fnal.gov:8200/ ... succeeded
Storing bearer token in /tmp/bt_token_accel_MARS_4395
Authentication needed for accel_MARS
Attempting OIDC authentication with https://htvaultprod.fnal.gov:8200/

Complete the authentication at:
    https://cilogon.org/device/?user_code=7QF-QWV-ZTC
No web open command defined, please copy/paste the above to any web browser
Waiting for response in web browser
Saving credkey to /nashome/b/boyd/.config/htgettoken/credkey-accel-MARS
Saving refresh token ... failed

htgettoken: Refresh token storage to https://htvaultprod.fnal.gov:8200/ failed: HTTPError: HTTP Error 403: Forbidden: 1 error occurred:
        * permission denied


condor_vault_storer: htgettoken failed
  More details might be available by running
    condor_vault_storer -v "accel_MARS" 

ERROR: (0) invoking /usr/bin/condor_vault_storer
Submitting job(s)
```

shreyb:
Note that this is from `/usr/bin/condor_submit`, not jobsub_submit, so it might be a bit tougher to tease out.

marcmengel:
We could maybe add a "...failed" afterwards if condor_submit is unhappy?

PR #231, opened 2023-01-20
print error on failed condor_submit status

marcmengel:


marcmengel:
This is an attempt to address #228 -- we don't suppress the "Submitting jobs" message, but we follow it 
with an error message that the submit failed.

shreyb (approved):
LGTM

shreyb:
Closes #228

PR #241, opened 2023-01-23
Two retry bugs:  Fixed retry logic if RCDS API call fails; Alert user and exit if RCDS tarball cannot be found after polling for it.

shreyb:


shreyb:
Thanks to @vitodb  for uncovering these bugs in his tests.

marcmengel (approved):
Looks good.

PR #250, opened 2023-01-27
make directories for pnfs dropbox case

marcmengel:
Don't know how I missed this...  but this works much better now.


shreyb (approved):
LGTM.  Does this pass unit tests?

shreyb:
Marc confirmed offline that this passes unit tests.  Merging.

PR #252, opened 2023-01-30
Submit transfer tests

marcmengel:
Integration tests for -f, -f dropbox://, --tar_file_name tardir://, etc.


shreyb (approved):
Looks good, with the caveat that three of our tests (do_tarballs{1...3} will now fail because we are observing the correct behavior, where we try to do `fake_ifdh.mkdir_p` the various fermilab directories.  This will get fixed when we update the tokens.

shreyb:
Now the fermilab bit works, but PNFS dropbox upload for non-fermilab VOs will  fail on any node that doesn't have the areas mounted NFS4, until the tokens are fixed.  Hopefully that'll be tomorrow.

shreyb (approved):

Issue #147, opened 2022-10-13
Method/service for fetching job logs over the web


retzkek:
Currently jobsub server makes job logs available to authenticated users over https, serving them from the local sandbox (which is where it submits jobs from and condor puts output into) on the schedd, with a cron job to clean those up after 30 days. This service is used by landscape jobview to provide a central SSO-authenticated method to access these logs.

Now, with jobsub_lite, the "sandbox" is on the user's submit machine, with input and output only available in the spool dir on the schedd, retrievable with `jobsub_fetchlog` (which uses `condor_transfer_data`), until the job is removed from the queue (either manually or by configuration). We have a couple options to provide web access for jobsub_lite jobs:

## Option 1: run server on schedd

Run a web service on each schedd that can serve files from the condor spool dirs, which would require tweaking some permissions knobs in condor.

### Pros
* Relatively straightforward to setup
* similar to old setup, simple integration with landscape jobview

### Cons
* condor includes stuff like credentials in the spool dir
* spool dir only exists while job is in queue, so we either keep jobs in the queue for 30 days (which has side-effects, like jobs not showing up in history until then) or reduce the availability window
* one more thing to maintain

## Option 2: dCache

Have condor put the output into a shared location, like dcache (or ceph via s3, coming soon(ish)). See https://htcondor.readthedocs.io/en/latest/users-manual/file-transfer.html?#file-transfer-using-a-url

### Pros
* Leverages existing services
* Files available after job leaves queue, cleaned up by dCache
* less things to maintain

### Cons 
* May require a custom file transfer plugin (or maybe http wil "just work"?)
* Need to verify dCache areas all set up for everyone in every group

### Steps

- [X] create area in scratch dCache for logs - RITM1589453 
- [x] add scope to tokens to write to the dCache area - RITM1602974
- [x] #255
- [x] job data for jobsub_lite jobs in Lens (currently expects to find "completed" jobs in history, not true now)
- [x] jobview fetch output from dCache

shreyb:
I think you've covered all the reasonable options, and in good detail.  Which one are you leaning towards?

retzkek:
I very much prefer Option 2, it's keeping with the "stick to condor" ethos of this project, and will reduce future maintenance and operation burden over Option 1 (as much as I would enjoy developing a new service). Unfortunately it doesn't "just work" in testing so far, our Vault tokens are formatted differently than the file transfer plugin expects. Should be a simple fix, we'll see what condor devs say.

[...]

PR #255, opened 2023-02-01
Transfer job sandbox to dcache

retzkek:
This is the first part of addressing #147. It modifies the wrapper script to transfer the truncated output (stdout and stderr) to dCache, as well as the job executable and jdf from the process 0. 

New/modified template variables:
* `date` - changed to be only the date: `%Y_%m_%d`
* `datetime` - what was previously `date`, including the time: `%Y_%m_%d_%H%M%S`
* `outurl` - base URL for the job sandbox in dCache, which is  `$JOBSUB_OUTPUT_URL/{{date}}/{{uuid}}`

New/modified classad attributes;
* `JobsubOutputURL`
* `JobsubUUID`

New  environment variable `JOBSUB_OUTPUT_URL`; if not set submission will continue without the transfers being included. A separate PR will be made with the config change.

~TODO: only simple templates are currently changed, still need to add this to dag templates.~


retzkek:
I believe this is ready, would appreciate a review. I've rebased onto the latest master.

templates/dataset_dag/dagbegin.cmd
```diff
@@ -4,7 +4,7 @@ universe           = vanilla
 executable         = sambegin.sh
 arguments          =
 
-{% set filebase %}sambegin.$(Cluster)$(Process){% endset %}
+{% set filebase %}sambegin.$(Cluster).$(Process){% endset %}
```
shreyb, line 7: Thanks for fixing this on the way.

templates/dataset_dag/sambegin.sh
```diff
@@ -50,16 +91,17 @@ chmod +x ${JSB_TMP}/ifdh.sh
...
 #EXPERIMENT=$1
 #DEFN=$2
 #PRJ_NAME=$3
 #GRID_USER=$4
 
 export JSB_TMP=$_CONDOR_SCRATCH_DIR/jsb_tmp
 mkdir -p $JSB_TMP
+redirect_output_start
 setup_ifdh_env
+echo `date` BEGIN executing /fife/local/scratch/uploads/fermilab/mengel/2020-01-14_162931.318256_6730/fife_wrap_20200114_162957_3316054.sambegin.sh
```
shreyb, line 103: This isn't just for you, Kevin, but also for @marcmengel .  Why are we saying here that we're starting a script that's in Marc's area...on /fife/local?  This seems very wrong...and possibly a holdover from testing.  Please correct me if I'm wrong.

templates/dataset_dag/samend.sh
```diff
@@ -48,11 +89,13 @@ chmod +x ${JSB_TMP}/ifdh.sh
 }
 
 
-echo `date` BEGIN executing /fife/local/scratch/uploads/fermilab/mengel/2020-01-14_162931.318256_6730/fife_wrap_20200114_162957_3316054.samend.sh
->&2 echo `date` BEGIN executing /fife/local/scratch/uploads/fermilab/mengel/2020-01-14_162931.318256_6730/fife_wrap_20200114_162957_3316054.samend.sh
 export JSB_TMP=$_CONDOR_SCRATCH_DIR/jsb_tmp
 mkdir -p $JSB_TMP
+redirect_output_start
 setup_ifdh_env
+echo `date` BEGIN executing /fife/local/scratch/uploads/fermilab/mengel/2020-01-14_162931.318256_6730/fife_wrap_20200114_162957_3316054.samend.sh
```
shreyb, line 96: Same comment as sam_begin.sh

shreyb (changes requested):
This largely looks good.  I'm concerned about the echo statements talking about /fife/local scripts being executed.

templates/dataset_dag/sambegin.sh
retzkek: Well that's embarrassing that I moved those lines but didn't notice that, good catch. Probably a "generated" script got mixed up with the template at some point. We could probably just change it to "BEGIN executing sambegin.sh"?

shreyb: That sounds good.

retzkek: changed

templates/dataset_dag/samend.sh
retzkek: changed

shreyb (commented):
LGTM.  Not merging, since we're cutting a release in 24 hours.  I'll approve and merge after that.

shreyb (dismissed):

PR #260, opened 2023-02-02
Fix #259

### LINKED ISSUE #259: jobsub_submit --maxConcurrent crash
It's trying to use `varg["dest"]` which is not defined; I'm not exactly sure what that should be?

```
$ jobsub_submit -G fermilab -n -N 1000 --maxConcurrent 10 file:///bin/true
  File "/opt/jobsub_lite/bin/jobsub_submit", line 309, in <module>
    main()
  File "/opt/jobsub_lite/bin/jobsub_submit", line 290, in main
    render_files(d1, varg, submitdir, dlist=[d1, d2, varg["dest"]])
KeyError: 'dest'
```

marcmengel:
Revisitng --maxConcurrent.   Found that we were missing template files from templates/maxconcurrent_dag (maybe never got "git add"-ed?) and a typo in the jobsub_submit code branch.    Further found that when we turned on the bits, it wasn't actually restricting number of jobs putting DAGMAN_MAX_SUBMITTED_JOBS in the config file, and we had to give a `-maxjobs n` argument to condor_dagman instead (I assume that that config value gets read by condor_submit_dag and it passes it with -maxjobs).   Looks like calling dagman with` -maxjobs 0 ` is fine when you don't have a limit, so not much fancy needed in the template.
Also added integration tests for it.


marcmengel:
Running integration tests... (now submits a fair number of jobs...)

marcmengel:
So integration tests passed except for not being able to put a tarfile in /pnfs/dune/resilient/jobsub_stage/... which I don't think is jobsub's problem.

shreyb (approved):

shreyb:
LGTM

PR #263, opened 2023-02-03
Changed disk default to 35GB to be inline with jobsub_client default

### LINKED ISSUE #262: --disk default is 100MB
https://github.com/marcmengel/jobsub_lite/blob/d79e3fe5aeb3a538e21e0568a4b8b39940419107/lib/get_parser.py#L168

@hschellman pointed out in INC000001146446 that the default --disk value is 100MB, which seems VERY low.  This was actually set in this commit 9c69e9d621a, so we should discuss why that value, and if it should indeed be changed, what value it should be.

We mention in the old jobsub_client documentation [here](https://cdcvs.fnal.gov/redmine/projects/jobsub/wiki/Getting_Started_with_the_Jobsub_Client) that the default disk value is 35GB.  I believe it should be that here.  We should similarly set all the other defaults to the same values as jobsub_client if they're already not there.

shreyb:
Fixes #262 

Unit tests pass.

Also, with the following jobsub_submit command:

```$ jobsub_submit -G fermilab -n file:///usr/bin/true```

We get the following request disk in the submit file:
```
$ grep -i disk  /home/sbhat/.cache/jobsub_lite/js_2023_02_03_035632_9fe6ea25-f37d-4135-9de8-72a5deec130d/simple.cmd 
request_disk = 36700160.0KB
```

So that works.

### DIFF
--- a/lib/get_parser.py
+++ b/lib/get_parser.py
@@ -165,7 +165,7 @@ def get_parser() -> argparse.ArgumentParser:
         " If UNITS is not specified default is 'KB' (a typo in earlier"
         " versions said that default was 'MB', this was wrong)."
         " Allowed values for UNITS are 'KB','MB','GB', and 'TB'",
-        default="100MB",
+        default="35GB",
     )
     parser.add_argument(
         "-d",


marcmengel (approved):
Looks good.

PR #266, opened 2023-02-07
Set role in args dict from return value of fake_ifdh.getRole

shreyb:
We were reading the role in from the default role file, but not actually setting it in a way that the template could see.  This fixes that issue.

### DIFF
--- a/lib/creds.py
+++ b/lib/creds.py
@@ -27,6 +27,7 @@ def get_creds(args: Dict[str, Any] = {}) -> Tuple[str, str]:
     """
 
     role = fake_ifdh.getRole(args.get("role", None))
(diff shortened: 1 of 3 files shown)


shreyb:
Ran all unit tests, and following manual tests:

Production:
1. --role production:  57125324.0@jobsub01.fnal.gov 
2. no role specified (should pick up from default role file): 57125325.0@jobsub01.fnal.gov 

Analysis: 57125326.0@jobsub01.fnal.gov

For all these cases, I did a dry-run where I checked that the .cmd file had the right oauth_services line.

marcmengel (approved):
Looks good!

PR #267, opened 2023-02-07
permissions fun

marcmengel:
Mimic old jobsub and make -f drobpox: files world readable before tar-ing them up (then put the permissions back).

### DIFF
--- a/lib/tarfiles.py
+++ b/lib/tarfiles.py
@@ -139,9 +139,19 @@ def do_tarballs(args: argparse.Namespace) -> None:
                 )
 
             if args.use_dropbox == "cvmfs" or args.use_dropbox is None:
+                # make sure they'll be able to read it, etc.
+                savemode = os.stat(pfn).st_mode
+                try:
+                    os.chmod(pfn, 0o755)
+                except OSError:
+                    pass
                 tarfile = tar_up(
                     os.path.dirname(pfn), "/dev/null", os.path.basename(pfn)
                 )
+                try:
+                    os.chmod(pfn, savemode)
+                except OSError:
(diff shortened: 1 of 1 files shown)


shreyb (changes requested):
Do we also need to do this for the --tar-file-name case?

marcmengel:
Well, in the general case, we would have to unpack the tarfile, change the file modes, and repack it... but I guess we *could*... It does look like the old jobsub does do that if it builds the tarfile (the tardir: case).  However the method they use of changing each file mode, adding it to the tarfile, then putting it back would require us to recode how we build the tarfiles, and use the python tarfile module to build the tarfile, and not external tar.  On the other hand, if we unpack the tarfile in $TMPDIR, chmod everything, and repack it, that would work in all cases...

marcmengel:
Hm... it does look like, with the tarfile module, we could just copy the tarfile and change the perms without really unpacking it into a directory... Let me give that a whirl..

marcmengel:
Ooh.. that's even almost pretty:

```
def tarchmod(tfn):
    ofn = tfn + '.o.bz2'
    with tarfile.open(tfn, 'r|*') as fin, tarfile.open(ofn, 'w|bz2')as fout:
        ti = fin.next()
        while ti:
            st = fin.extractfile(ti)
            ti.mode = 0o755
            fout.addfile(ti,st)
            ti = fin.next()

```

shreyb:
Discussed offline.  We'll merge this as is now, since it fixes the immediate bug, and then pursue a more elegant way of handling tarfile permissions in #268.

shreyb (approved):

PR #272, opened 2023-02-09
Put quotes around --os argument

### LINKED ISSUE #271: --os flag result is missing double-quotes in template
Thanks to @StevenCTimm for this bug report:

In our templates, the value for the `--os` flag should be double-quoted, if it's not `None`.

https://github.com/marcmengel/jobsub_lite/blob/0576298895c59dd143153d73332d1d34e77903fa/templates/simple/simple.cmd#L29

In the example above, we can see, that, for example, if a user types:

`jobsub_submit -G experiment --os SL7 file:///path/to/file`

the resultant .cmd file entry is:

`+DesiredOS=SL7`.

It should be 

`+DesiredOS="SL7"`

This fix should be applied to all applicable templates.

shreyb:
Closes #271 

One thing for reviewers to check.  Is it REALLY only set in the one place?  I couldn't find any other templates outside of simple.cmd where we set the DesiredOS classad, but should it be in the other templates too?


marcmengel:
Yes.  We don't set Desired_OS on the submits for dagmans because they run on the scheduler..

shreyb:
That totally makes sense - slipped my mind.  Thanks for clarifying that.

marcmengel (approved):
Looks good.

PR #273, opened 2023-02-09
make -f files executable

marcmengel:
In case they're uploading a script...

### DIFF
--- a/templates/simple/simple.sh
+++ b/templates/simple/simple.sh
@@ -210,7 +210,7 @@ export PATH="${PATH}:."
 # -f files for input
 {%for fname in input_file%}
 ${JSB_TMP}/ifdh.sh cp -D {{fname}} ${CONDOR_DIR_INPUT}
-chmod u+x ${CONDOR_DIR_INPUT}/{{fname}}
+chmod u+x ${CONDOR_DIR_INPUT}/{{fname|basename}}
 {%endfor%}
 
 # --tar_file_name for input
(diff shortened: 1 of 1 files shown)


shreyb (commented):
LGTM, but I'm going to test it before approving.

templates/simple/simple.sh
```diff
@@ -210,6 +210,7 @@ export PATH="${PATH}:."
 # -f files for input
 {%for fname in input_file%}
 ${JSB_TMP}/ifdh.sh cp -D {{fname}} ${CONDOR_DIR_INPUT}
+chmod u+x ${CONDOR_DIR_INPUT}/{{fname}}
```
shreyb, line 213: So there's a mistake here.  As is, we'd be trying to do something like 

`chmod u+x ${CONDOR_DIR_INPUT}/cvmfs/fifeuser1.opensciencegrid.org/sw/mu2e/4735aadedec8938e9ba7d03829af68a5d03429667f8dfcce5d8e95df29929dff/myexecutable`, where what we want is

`chmod u+x ${CONDOR_DIR_INPUT}/myexecutable`

I think this line needs to be:

`chmod u+x ${CONDOR_DIR_INPUT}/{{fname|basename}}`

Does that sound right?

shreyb (changes requested):

templates/simple/simple.sh
marcmengel: Yep.

shreyb (approved):
LGTM.

PR #275, opened 2023-02-10
collapse identical (except for worker job number) DAG stages

marcmengel:
...into same .cmd and wrapper script.   Reduces project.py submission files considerably.


marcmengel:
Note: view changes with "Hide whitespace" on in the settings, there's a big block indented with very small changes in it...

lib/dagnabbit.py
```diff
@@ -133,50 +136,80 @@ def parse_dagnabbit(
...
+                        )
+                    with open(
+                        os.path.join(dest, f"{name}.sh"), "w", encoding="UTF-8"
+                    ) as csf:
+                        csf.write(
+                            jinja_env.get_template("simple.sh").render(**thesevalues)
+                        )
+                    of.write(f"\nJOB {name} {name}.cmd\n")
+                    of.write(
+                        f'VARS {name} JOBSUBJOBSECTION="{count}" CM2="{count-2}" CM2="{count-1}" nodename="$(JOB)"\n'
```
shreyb, line 211: Shouldn't this be `CM2="{count -2}" CM1="{count-1}"` like above (line 153)?  I see two `CM2`'s in this line.

shreyb (changes requested):
Just the one comment on line 211.  

Generally, how sure are we that this won't break existing workflows that, thus far, have worked with jobsub_lite?  Are our unit and integration tests for the DAGs good enough to give us confidence, and/or is there a unit test we can write that tests this project.py case?

lib/dagnabbit.py
marcmengel: Ack!  yes.

marcmengel:
Also added a test case (based on a project.py dag from Herb) that checks that we collapsed stages.

tests/dagnabbit/dagTestH
```diff
@@ -0,0 +1,40 @@
+<serial>
+
+<parallel>
+
+jobsub_submit --group=uboone --resource-provides=usage_model=DEDICATED,OPPORTUNISTIC --lines='+SingularityImage=\"/cvmfs/singularity.opensciencegrid.org/fermilab/fnal-wn-sl7:latest\"' --expected-lifetime=3h --memory=1000MB --append_condor_requirements='(TARGET.HAS_CVMFS_uboone_opensciencegrid_org==true)' --append_condor_requirements='(TARGET.HAS_CVMFS_uboone_osgstorage_org==true)' --append_condor_requirements='(TARGET.HAS_SINGULARITY=?=true)' file://jobA.sh  --max_files 10  --prestage_fraction 0.750000  --logdir /pnfs/uboone/scratch/users/greenlee/v08_00_00_55/swizzle2/swizzle2
```
shreyb, line 5: Do we want to have tests that are tied to Herb's account?  I'm fine keeping them in as a temporary solution to make sure we get good tests, but we should generalize it as soon as possible afterwards.

shreyb (approved):
Approved, but please look at the one comment.  If the change is non-trivial, we can address that later, but we definitely need to address it and not let it fall off our radar.  

Also, please don't merge this as of yet.  I'll merge any open PRs that aren't bugfixes tomorrow morning, after 1.0.2 is deployed.

tests/dagnabbit/dagTestH
marcmengel: Good catch.  I switched it to $USER

shreyb (approved):
LGTM.  I'll hold off on merging this until tomorrow, after 1.0.2 is released.

Issue #277, opened 2023-02-13
Change default disk request to 10 GB to match current default in jobsub_client


kherner:
In Issue 263 the default disk request went to 35 GB, ostensibly to match the (old) jobsub_client default. The current default is in fact 10 GB:

>  RequestDisk = ifThenElse(JobUniverse != 7,10000000,16000)

It was 35 GB a while back but it was changed a few years ago because we had seen issues with completely filling up newer FermiGrid worker nodes as they had smaller scratch disks.

PR #278, opened 2023-02-13
Change disk default to 10GB

shreyb:
Thanks to @kherner for pointing this out.

We were using an outdated value for our default.  This was the document, which has since been updated:

https://cdcvs.fnal.gov/redmine/projects/fife/wiki/Information_about_job_submission_to_OSG_sites


shreyb:
Passes unit tests and a simple submission check of the .cmd file.

shreyb:
Closes #277

marcmengel:
So I'm not sure if we shouldn't have default=None, to let the schedd/glideins pick the default(?).

marcmengel (approved):
Looks good to me.

Issue #283, opened 2023-02-17
Feature request:  Get individual section logs back from big*** jobsub jobs


hschellman:
Use case:

Rogue user submits a job with 10000 individual jobs.  

Some of them fail quickly, others are still running.   

right now a request for logs in this case returns a broken tarball that is partially openable.  But even if they all finish, the log file that comes back is immense and hard to parse. 

request is for the ability to request the logs for a particular section (or whatever it is called) so one can see what is happening quickly and definitely before the whole group of jobs ends.

PR #286, opened 2023-02-20
add --partial argument to jobsub_fetchlog

retzkek:
We (I) inadvertently left it out of the rewrite. condor_transfer_data will abide whatever the constraint is, we were forcing it to be the cluster.

### DIFF
--- a/bin/jobsub_fetchlog
+++ b/bin/jobsub_fetchlog
@@ -70,7 +70,7 @@ def main():
     parser.add_argument(
         "--partial",
         action="store_true",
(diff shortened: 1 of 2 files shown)


retzkek:
```download only the stdout and stderr for this jobid, not the entire sandbox```

I copied the help text from jobsub_client, but it's not technically true, since the user will also get the cmd, scripts, and event log files. Does it matter enough to change it?

retzkek:
I forgot to mention this will fix #283.

shreyb (approved):
LGTM

PR #292, opened 2023-02-21
Use a makefile to build RPM rather than script

shreyb:
This basically replicates the build script, but it hopefully makes it more obvious how to build a release.  

This does NOT need to go into 1.1, so even if it's approved, I'd request that we delay merging it until after 1.1 is released.


shreyb:
I should add, I tested this by building a fake "1.0.2a" release and installed the resultant RPM on my dev machine.  If this is approved, I'll go ahead and update the documentation as well.

marcmengel (commented):
Looks lovely.

marcmengel (approved):

shreyb:
Will merge after 1.1 release is deployed.  No reason to rush this in.

PR #293, opened 2023-02-21
add JobsubOutputURL and JobsubUUID to dag sub files

retzkek:
These classad attributes were inadvertently left out of the DAGs, they're needed for jobview to find the log path on dcache.


shreyb (approved):
LGTM.

PR #294, opened 2023-02-21
Go back to the assumption that users have decode_token.sh on their PATH.  Added bin dir to PATH for testing

shreyb:
In #280 , we broke the very esoteric use case of a python script calling jobsub_submit after doing an `os.chdir`.  So we go back to the assumption that `decode_token.sh` is in the `PATH`, and add it in for the testing case.


shreyb:
This passes all unit tests.

marcmengel (approved):
Looks good, although I'm not sure why we're adding the bin directory to sys.path/PYTHONPATH, but it shouldn't hurt.

PR #295, opened 2023-02-21
Sanitize --constraint flag before passing argument to condor

### LINKED ISSUE #289: --constraint flag should allow for --constraint="constraint_value" syntax
Currently, we pass the `--constraint` value in `jobsub_cmd` straight to condor when we run jobsub_q, jobsub_rm, etc.  The problem is that condor doesn't support `-constraint="my_fake_constraint"` syntax, and so when we pass something like 

`jobsub_q -G mygroup --constraint="my_fake_constraint"`

we get an error, because condor doesn't know what to do.  It's a special case, but a commonly-used one, so we should take care to pass the correct constraint flag if it's given on the command line by the user.

shreyb:
Closes #289

### DIFF
--- a/bin/jobsub_cmd
+++ b/bin/jobsub_cmd
@@ -85,6 +85,9 @@ def main() -> None:
         passthru.append(arglist.jobid)
     if arglist.user:
         passthru.append(arglist.user)
+    # If they gave us --constraint, sanitize it and add to passthru
+    if arglist.constraint:
+        passthru.extend(["-constraint", arglist.constraint])
 
     if os.environ.get("GROUP", None) is None:
         raise SystemExit(f"{sys.argv[0]} needs -G group or $GROUP in the environment.")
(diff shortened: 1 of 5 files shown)


shreyb:
Note:  I cherry-picked one of these commits for another PR.  So this PR will change before I mark it ready for review.  Also, I want to add an integration/unit test.

shreyb:
Checked with the following commands:

```
$ jobsub_q -G fermilab --constraint='Owner=="dunepro"' --jobid=6113.0@jobsubdevgpvm01.fnal.gov 
$ jobsub_q -G fermilab --constraint 'Owner=="sbhat"' --jobid=6113.0@jobsubdevgpvm01.fnal.gov 
$ jobsub_q -G fermilab --constraint='Owner=="sbhat"' --jobid=6113.0@jobsubdevgpvm01.fnal.gov 
$ jobsub_q -G fermilab --constraint 'Owner=="dunepro"' --jobid=6113.0@jobsubdevgpvm01.fnal.gov 
```

That job was one submitted by me, so the dunepro checks were invalid, but still should get parsed.  The sbhat checks should be valid, and of course get parsed.

shreyb:
All unit tests pass.

marcmengel (approved):
Excellent!  I like the tests.

PR #298, opened 2023-02-22
Move cmd and script transfers to end of job

retzkek:
Transfer info was showing up at the top of stderr, which could confuse and anger users. Here they won't show up in the web logs at all, but could still be seen in the spool logs for debugging. Downside is if the job is violently killed such that the postscript functions don't get to run, they won't get transferred.

Also includes a couple comments regarding changing log output url and/or path for posterity.


shreyb (approved):
LGTM.  Thanks for noting that bit about if the job gets killed in an unusual way, we won't see this go through.

PR #303, opened 2023-02-23
Fixes to fixes

marcmengel:
This fixes the tarball problems introduced with symlinks, adds LD_LIBRARY_PATH-proofing to rest of scripts in bin, and adds extra horizontal lines and whitespace to make the job_id  stand out.

### DIFF
--- a/bin/condor_submit
+++ b/bin/condor_submit
@@ -23,6 +23,11 @@ import sys
 import os
 import os.path
 import argparse
+
+if os.environ.get("LD_LIBRARY_PATH", ""):
+    del os.environ["LD_LIBRARY_PATH"]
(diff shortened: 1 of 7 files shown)


shreyb:
@marcmengel Love the highlighting.

I ran unit tests on this, including Marc's new one that has a symlink in the test tarball, and also ran the following (a pseudo-integration test):

```jobsub_submit -G fermilab --tar-file-name tardir:///home/sbhat/TestJobs/TestDir file:///usr/bin/sleep 300```

I can confirm that with current `master`, this fails, but with this PR checked out, it works.

Approved.

shreyb (approved):

PR #309, opened 2023-02-25
hard and softlink tar handling, modified tarfile in current directory

marcmengel:


marcmengel:
We should review this after a bit more testing.

lib/tarfiles.py
```diff
@@ -65,16 +65,16 @@ def __call__(self, r: requests.Request) -> requests.Request:
 def tarchmod(tfn: str) -> str:
     """copy a tarfile to a compressed tarfile changing modes of contents to 755"""
     os.environ["GZIP"] = "-n"
-    ofn = f"{tfn}.o.gz"
+    ofn = os.path.basename(f"{tfn}.o.gz")
```
shreyb, line 68: Do we want to also catch an exception in case the user tries to create the tarball in a read-only dir, and give a helpful error message?  Something like "jobsub_lite needs to create a temporary copy of the tarball you are attempting to upload, {tfn}, in the current working directory, but this directory is read-only.  Please change directories to somewhere you have permission to write to, and try resubmitting your job(s)."

What do you think?

shreyb (changes requested):
LGTM, but I have a single comment about maybe we should warn users if they're on a read-only FS.

lib/tarfiles.py
marcmengel: We should do that several places in here, then...  We could make one routine for it and call it in several places?   Do you think it is sufficient to just check if the current directory is writable?

shreyb: Yeah - I think that's sufficient for now.  @retzkek and I were talking offline about how we need to be catching errors in a lot of other places in the code, rather than spitting tracebacks at users.  We could add a ticket for tarfile errors, especially because it's a few layers down, and can be confusing for even us to figure out the issue.

shreyb:
Note:  This PR comes from ticket INC000001147543 in ServiceNow, where a mu2e user previously was unable to submit a job with a tarball, and still won't be able to after 1.1.1 is deployed.  This PR should fix that issue.

I took the user's tarball, and reproduced the error in 1.1.1, and have confirmed that with this PR's code checked out, the error disappears and jobs can be submitted.  We'll be reaching out to the user to test on Monday, as well as running our own tests, as @marcmengel mentioned above.

lib/tarfiles.py
marcmengel: Done.  I added a check_writable routine that we call before we do tarfile creation; it checks if the current directory is writable, and  it explains that it really wanted to write in the current directory but couldn't, and exits.

shreyb (commented):
LGTM.  We'll test it out more on Monday before approving/merging, as Marc mentioned.

PR #312, opened 2023-02-27
find schedds with collector.query() for DUNE global pool

marcmengel:
Quick stab at making dune global pool work again.  Limiting schedd's to ones that match our domain name. 
(This in theory could then work at bnl, as well... maybe?)

### DIFF
--- a/lib/condor.py
+++ b/lib/condor.py
@@ -19,6 +19,7 @@ import sys
 import glob
 import re
 import random
+import socket
 import subprocess
 from typing import Dict, List, Any, Tuple, Optional, Union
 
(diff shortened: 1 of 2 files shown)


marcmengel:
Okay, so now revisited, it does the whole business of picking a schedd list with one condor collector.query()
Then we aren't contacting the schedd directly, and we don't have problems with the offsite schedd (unless they 
set the IsJobsubLite flag on it, but that's another whole kettle of fish).

lib/condor.py
```diff
@@ -39,61 +39,20 @@ def get_schedd(vargs: Dict[str, Any]) -> classad.ClassAd:
...
     # pylint: disable-next=no-member
-    schedd_classads = coll.locateAll(htcondor.DaemonTypes.Schedd)
-
-    # locateAll gives a list of minimal classads... but
-    # need to directQuery for the full classads to check for
-    # SupportedVOList...
+    devnot = "" if vargs["devserver"] else "!"
+    schedds = coll.query(
+        htcondor.htcondor.AdTypes.Schedd,
+        constraint="isJobsubLite=?=true"
```
shreyb, line 45: I don't know if this matters, but I believe`IsJobsubLite` should be capitalized.

shreyb: Certainly the classad is:

```[rexbatch@dunegpcoll01 ~]$ condor_status -schedd -l | grep -i isjobsublite
IsJobsubLite = true
IsJobsubLite = true
IsJobsubLite = false
```

lib/condor.py
```diff
@@ -39,61 +39,20 @@ def get_schedd(vargs: Dict[str, Any]) -> classad.ClassAd:
...
+        " && "
+        f"""STRINGLISTIMEMBER("{vargs['group']}", SupportedVOList)"""
+        " && "
+        "InDownTime != true"
+        " && "
+        f'{devnot} regexp(".*dev.*", Machine)',
```
shreyb, line 51: Love it.

shreyb (changes requested):
Just the one bit about capitalization of `IsJobsubLite`

lib/condor.py
marcmengel: Yes.  Missed that, thanks!

shreyb (approved):
LGTM.  I ran all unit tests, and the following:

1. jobsub_submit -G fermilab --verbose=2  file:///usr/bin/sleep 300
2. jobsub_submit -G fermilab --verbose=2  --devserver file:///usr/bin/sleep 300
3. export _condor_COLLECTOR_HOST=dunegpcoll02.fnal.gov; jobsub_submit -G dune --verbose=2  file:///usr/bin/sleep 300
4. jobsub_submit_dag -G dune --verbose=2  file:///home/sbhat/TestDAG/testdag2.dag
5. unset _condor_COLLECTOR_HOST; jobsub_submit -G dune --verbose=2  file:///usr/bin/sleep 300


All submitted jobs successfully, and to the correct pools.  Will wait on further tests from @marcmengel and @kherner tomorrow morning before we merge this PR into 1.1.2.

shreyb (approved):

shreyb (approved):

PR #316, opened 2023-02-28
dagman crunching patch for values ending up in the shell script

marcmengel:
Yet another patch on the DAG-crunching, for when the count gets passed through into the wrapper script on the command arguments; where we need to convert the DAG/submit parameters to environment variables to use them.


shreyb (commented):
Is there a unit test/integration test that proves this fixes the issue? 

Aside from that, LGTM.

PR #317, opened 2023-02-28
only do $(CM1) $(CM2)  stuff at end of command line to aviod causing other problems

marcmengel:
For problems Lisa was looking into where `jobsub_submit -N 1` was becoming` jobsub_submit -N $(CM1) `and hosing up parsing


shreyb (approved):
LGTM.  Passes all unit tests, including ones based very closely on @hgreenlee and @goodenou use cases.

Issue #320, opened 2023-03-03
Let DAG nodes inherit tarball path


vitodb:
For DAG submissions in order to let each node to be aware of the RCDS path of a tarball it is needed to use the `--tar_file_name dropbox://` option for each DAG node.
I'm wondering if it could be possible to implement a feature that allows to use the `--tar_file_name dropbox://` option for the DAG submission and let all DAG nodes inherit the associated env var `INPUT_TAR_DIR_LOCAL` and other related stuff, so avoiding the need for each DAG node to (try to) upload the tarball to RCDS.

marcmengel:
That was *supposed* to work already... I have reproduced this and fixed it; pull request forthcoming after tests...  The other thing I might still put in is if you have a --tar-file-name on the main command line and one in the dag lines, you should get *both* -- we should merge the lists, not override them

PR #321, opened 2023-03-03
Tarfile dag

marcmengel:
This addresses an issue Vito brought up, in #320, but also rubbed my nose in the fact that we weren't doing combinations of, say -e arguments on the command line versus the dagnabbit file properly -- if the dagnabbit file specified -e anything, that overrode *all* -e arguments from the command line, etc. when we should merge them.

Also switched the tarfile regeneration to make .bz2 files rather than .gz ones, as we had the random-hash due to current time showing up in those .gz tarballs, even with GZIP=-n set in the environment.  Now we get consistent hashes for tarfiles again.

### DIFF
--- a/lib/tarfiles.py
+++ b/lib/tarfiles.py
@@ -73,10 +73,9 @@ def check_we_can_write() -> None:
 
 def tarchmod(tfn: str) -> str:
     """copy a tarfile to a compressed tarfile changing modes of contents to 755"""
-    os.environ["GZIP"] = "-n"
-    ofn = os.path.basename(f"{tfn}.o.gz")
+    ofn = os.path.basename(f"{tfn}.o.bz2")
     check_we_can_write()
-    with tarfile_mod.open(tfn, "r|*") as fin, tarfile_mod.open(ofn, "w|gz") as fout:
+    with tarfile_mod.open(tfn, "r|*") as fin, tarfile_mod.open(ofn, "w|bz2") as fout:
         ti = fin.next()
         while ti:
             if ti.type in (tarfile_mod.SYMTYPE, tarfile_mod.LNKTYPE):
--- a/lib/tarfiles.py
+++ b/lib/tarfiles.py
@@ -73,7 +73,7 @@ def check_we_can_write() -> None:
 
 def tarchmod(tfn: str) -> str:
     """copy a tarfile to a compressed tarfile changing modes of contents to 755"""
-    ofn = os.path.basename(f"{tfn}.o.bz2")
+    ofn = os.path.basename(f"{tfn}.tbz2")
     check_we_can_write()
     with tarfile_mod.open(tfn, "r|*") as fin, tarfile_mod.open(ofn, "w|bz2") as fout:
         ti = fin.next()
@@ -93,7 +93,7 @@ def tar_up(directory: str, excludes: str, file: str = ".") -> str:
     """build directory.tar from path/to/directory"""
     if not directory:
         directory = "."
-    tarfile = os.path.basename(f"{directory}.tar.gz")
+    tarfile = os.path.basename(f"{directory}.tgz")
     check_we_can_write()
     if not excludes:
         excludes = os.path.dirname(__file__) + "/../etc/excludes"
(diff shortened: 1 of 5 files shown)


shreyb (changes requested):
Could you please update your config submodule commit?  You're one commit behind, I believe (it should be [e0110fd](https://github.com/marcmengel/jobsub_lite_config/tree/e0110fd00b531918ce81674393cea8cde50ca2ff))

lib/tarfiles.py
```diff
@@ -73,10 +73,9 @@ def check_we_can_write() -> None:
 
 def tarchmod(tfn: str) -> str:
     """copy a tarfile to a compressed tarfile changing modes of contents to 755"""
-    os.environ["GZIP"] = "-n"
-    ofn = os.path.basename(f"{tfn}.o.gz")
+    ofn = os.path.basename(f"{tfn}.o.bz2")
     check_we_can_write()
-    with tarfile_mod.open(tfn, "r|*") as fin, tarfile_mod.open(ofn, "w|gz") as fout:
+    with tarfile_mod.open(tfn, "r|*") as fin, tarfile_mod.open(ofn, "w|bz2") as fout:
```
shreyb, line 78: The idea here is that bz2 gives us hash reproducibility independent of time, and gz does not?

marcmengel: Exactly.  I *thought* it would get fixed with the same GZIP=-n in the environment that works for actual /usr/bin/tar, but no such luck.

shreyb: Yeah - I believe this is what jobsub_client used to do, so I'm surprised it doesn't work here.  What's the effect on currently-uploaded tarballs?  The hashes will all be different than before, so basically every single new submission using dropbox:// or tardir:// will upload a new tarball to RCDS, right?

shreyb (commented):
While we're looking at tarballs, do we want to add to line 214/215 of tarfiles.py that if bz2 tarballs should get their extensions stripped like .tar and .tgz?

If you want to split that into a separate issue, since it's tangentially related, that's fine with me.

shreyb (approved):

marcmengel:
Well, once, yes.  But right now its re-uploading them every time, so once is an improvement.   Note that none of this takes much space on the cvmfs side, since cvmfs collapses duplicate files (i.e. of the unpacked tarfiles)  internally.

shreyb:
Sorry - I'm confused.  Are you referring to this PR, or #323?

marcmengel:
I was replying to "...What's the effect on currently-uploaded tarballs? The hashes will all be different than before, so basically every single new submission using dropbox:// or tardir:// will upload a new tarball to RCDS, right"

shreyb:
Ah - my mistake!  Thanks for that clarification.  Anyway, I'm good with this PR.  Do you want to make the change on line 214 of lib/tarfiles.py or should we handle that separately?

marcmengel:
Looking at it...

marcmengel:
Okay, did a little more cleanup of tarfile suffxes all around... what do you think?

lib/tarfiles.py
```diff
@@ -73,10 +73,9 @@ def check_we_can_write() -> None:
 
 def tarchmod(tfn: str) -> str:
     """copy a tarfile to a compressed tarfile changing modes of contents to 755"""
-    os.environ["GZIP"] = "-n"
-    ofn = os.path.basename(f"{tfn}.o.gz")
+    ofn = os.path.basename(f"{tfn}.tbz2")
```
shreyb, line 76: BTW - do we know that RCDS plays nicely with tbz2 files?  I don't see why it wouldn't, but the integration tests went OK with this change?

shreyb: Besides wanting to double-check this, this PR LGTM.

marcmengel: I'll rerun them. I can't recall now if I last ran them before or after this change.. sigh.

marcmengel: (I mean , with the suffix tbz2;  it's done the gzipped tarballs already...)

shreyb: Did these tests work?  If so, I'll go ahead and merge.

marcmengel: Sigh... My tests on friday got stuck on a job that just wouldn't come out of Idle... I'm re-running.

marcmengel: Okay, found one bug so far... Still need to figure out what test is having the other problem...  It's making a very funky dag submission which keeps getting restarted and exiting 255 in the joblog; and doesn't seem to have the dag_wrapper.sh set as the executable, and.?!?!?

marcmengel:
Okay, I think this is good to merge now; sorry for the delay, running jobs for unit tests is sometimes Really Slow...

PR #322, opened 2023-03-03
transfer submission files at submit time, not from the job

### LINKED ISSUE #311: Transfer submission files (script, JDF) to dCache at submission time?
This is for discussion: right now we have to transfer the description file to the job, in order for it to get copied out to dCache. We could copy it and the script there right at submission time from jobsub_submit.

Pros:
* creates output destination, users won't get 404 error trying to see what files are there (currently it won't be created until the first job has finished)
* don't have to copy the JDF in input_files

Cons:
* more complexity in jobsub_submit, including IFDH (?) dependency... or maybe we just do straight webDAV calls since it's a very well-defined transfer?

retzkek:
So don't even transfer them TO the job to begin with. This transfers every file in the submit dir, which I think is OK (like, we never expect there to be credentials in there?)

 Fixes #311


shreyb:
Looks like you need to rebase?

retzkek:
'tis done, twice since the first time I messed up. `rebase -i` (or really magit `r i`) is a lifesaver :)

shreyb:
As far as whether or not we expect credentials to be in the submit directory - I'm not sure we can make that assumption.  @vitodb would be a good person to ask this question to.  If anyone has a legitimate reason to do so, it'd probably be him.

vitodb:
Is here `submit directory` the actual directory where users run the `jobsub_submit` command?

retzkek:
@vitodb no sorry, "submit dir" means the directory that jobsub_submit creates for the submission in ~/.cache/. It renders the job description files and scripts in there, then calls condor_submit. I don't think there's a (supported) way for users to put other files in there at submit time, but I'm not 100% sure.

retzkek:
@marcmengel can you in particular look over the changes to all the template files and make sure I'm not being overzealous (or under) with the transfer_input and _output file lists?

vitodb:
@retzkek thanks to clarify. In my past test using `--no-submit` option, mostly for DAG submissions, to inspect what jobsub_lite was creating in the "submit dir", I didn't saw any credential bit.
The only "custom" file that I'm aware a user can get in there is through `file://`, which is supposed to be a script/executable. 
However having a word from @marcmengel would be better :)

marcmengel:
So there was a time, early on in the project, where we were assuming a shared filesystem, and I thought I needed the credentials in the submit directory, so there may be some vestige of that in comments, etc...
but it turns out you don't need them there.

bin/jobsub_submit
```diff
@@ -163,6 +164,29 @@ def do_dataset_defaults(varg: Dict[str, Any]) -> None:
...
+    try:
+        mkdir_p(dest_url)
+    except Exception as e:
+        print(
+            f"warning: error creating sandbox, web logs will not be available for this submission: {e}"
+        )
+        return
+    for f in os.listdir(src_dir):
+        try:
+            cp(os.path.join(src_dir, f), dest_url)
```
marcmengel, line 183: Could we make this `cp(os.path.join(src_dir, f), os.path.join(dest_url,f))  ` so we aren't relying on the check if the destintion is a directory behavior?

marcmengel (approved):
Looks good; I have a nit there about the cp(file,directory) we might want to change, but otherwise fine.

retzkek:
I just realized that my testing so far has been with a proxy present, and fake_ifdh won't even use a token (since gfal will only look for `BEARER_TOKEN`). So need to add some more logic there.

shreyb:
@marcmengel given Kevin's and your most recent comments, can I dismiss your review so we don't accidentally merge it before it's ready?

retzkek:
There now you're a reviewer too so don't approve it yet

marcmengel:
@retzkek wrote:
> I just realized that my testing so far has been with a proxy present, and fake_ifdh won't even use a token (since gfal will only look for `BEARER_TOKEN`). So need to add some more logic there.

Well, we can add the `[ -s ${BEARER_TOKEN_FILE:-/dev/null} ] && export BEARER_TOKEN=`cat ${BEARER_TOKEN_FILE}`;  ` bits to the gfal_clean_env string to get it to use a token; **but** we need to wait for that until the persistent areas are in the token scopes, otherwise we'll break` --use-pnfs-dropbox.`..

bin/jobsub_submit
retzkek: Good call. changed.

retzkek:
> but we need to wait for that until the persistent areas are in the token scopes, otherwise we'll break --use-pnfs-dropbox...

Ach! I was just about to push something to "fix" that. OK I guess we'll call that a TODO for when we get around to removing proxies and let gfal use the proxy that we currently get anyways.

shreyb:
> but we need to wait for that until the persistent areas are in the token scopes, otherwise we'll break 
> --use-pnfs- dropbox...

I thought that `--use-pnfs-dropbox` writes to `resilient`, and those scopes _are_ in the token...

shreyb (approved):
LGTM.  Any other comments, @marcmengel, before we merge?

PR #326, opened 2023-03-07
weight schedd selection by RecentDaemonCoreDutyCycle

marcmengel:
From my discussion with Nick... We can even up busy vs not busy schedds.

### DIFF
--- a/lib/condor.py
+++ b/lib/condor.py
@@ -54,7 +54,29 @@ def get_schedd(vargs: Dict[str, Any]) -> classad.ClassAd:
     if vargs.get("verbose", 0) > 1:
         print(f"post-query schedd classads: {schedds} ")
 
-    res = random.choice(schedds)
+        print("")
+
+    weighted_schedds = []
+    for s in schedds:
+        name = s.eval("Name")
+        rdcdc = s.eval("RecentDaemonCoreDutyCycle")
+
+        if vargs.get("verbose", 0) > 0:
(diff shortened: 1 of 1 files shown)


lib/condor.py
```diff
@@ -54,7 +54,29 @@ def get_schedd(vargs: Dict[str, Any]) -> classad.ClassAd:
...
+        weighted_schedds.append(s)
+        if rdcdc < 0.5:
+            weighted_schedds.append(s)
+        if rdcdc < 0.1:
+            weighted_schedds.append(s)
+        if rdcdc < 0.05:
+            weighted_schedds.append(s)
+            weighted_schedds.append(s)
+
+    res = random.choice(weighted_schedds)
```
shreyb, line 79: This is fine, but python 3.6 and above has a `random.choices` that allows you to weight your choices.  

https://docs.python.org/3/library/random.html#random.choices

I'll leave it up to you as to whether to use that or not, but it might come out a bit idiomatic/cleaner, especially the part where you compute the weights.

shreyb (changes requested):

lib/condor.py
marcmengel: Trying that...

marcmengel: Does this look better now @shreyb ?

shreyb (approved):
Thanks Marc - this looks great.

PR #327, opened 2023-03-09
need to put SOMETHING in transfer_output_files...

retzkek:
or condor transfers "everything" back to the schedd.


retzkek:
.empty_file is (and always has been) created by simple.sh. "small" bit of institutional knowledge that got lost.

shreyb (approved):

Issue #330, opened 2023-03-13
If --lines is specified with --singularity-image, we override --lines


shreyb:
This issue originated from the ServiceNow ticket INC000001148397, though this ticket doesn't address that issue (which is most likely user error)

jobsub_lite doesn't support specifying `--lines` with a `SingularityImage` classad, because its default for `--singularity-image` overrides that.  We need to allow for the following:

1.  If `--singularity-image` is specified and is NOT the default, use that.
2. If `--lines` is given with a `+SingularityImage` classad, and _that_ value is not the default, use that.
3. Else, use the default.

The assumption we're making here is that if you specify a non-default --singularity-image, that's what you want.

shreyb:
We also need to document this on the Differences wiki page.

shreyb:
Code changes made here: https://github.com/shreyb/jobsub_lite/tree/issue-330

Need to write unit tests.

shreyb:
Unit tests done, and the following no-submit tests, where I ran the indicated jobsub_submit line, and inspected the generated `simple.cmd` file:

Test 0a:  Only --singularity-image with non-default image - PASS

```$  ~/jobsub_lite/bin/jobsub_submit -G fermilab --singularity-image="/cvmfs/singularity.opensciencegrid.org/fermilab/fnal-wn-sl6:latest" -n file:///usr/bin/sleep 300

$ grep 'fake\|Singularity'  /home/sbhat/.cache/jobsub_lite/js_2023_03_15_015018_cd864e07-990b-4442-88b3-d4a699497aee/simple.cmd 
+SingularityImage="/cvmfs/singularity.opensciencegrid.org/fermilab/fnal-wn-sl6:latest"
```

Test 0b:  Only --no-singularity, --lines with SingularityImage - should strip SingularityImage out of lines - PASS

```
$  ~/jobsub_lite/bin/jobsub_submit -G fermilab --no-singularity  --lines '+SingularityImage=\\\"/cvmfs/singularity.opensciencegrid.org/fermilab/fnal-wn-sl5:latest\\\"' --lines '+JobsubTestClassad=\\\"thisisafakeclassad\\\"' -n file:///usr/bin/sleep 300
Warning:  --lines contains a SingularityImage specification but --no-singularity was also passed on command line. jobsub_lite will remove the --lines parameter that contains SingularityImage.

[...]

PR #331, opened 2023-03-14
Resolve --singularity-image and --lines with +SingularityImage conflicts

shreyb:
As described in #330 , we establish the following convention here with regards to the `--singularity_image` flag and `--lines` flag with `SingularityImage` specified:

1.    If --singularity-image is specified and is NOT the default, use that.
2.   If --lines is given with a +SingularityImage classad, and that value is not the default, use that.
3.   Else, use the default.

### DIFF
--- a/lib/get_parser.py
+++ b/lib/get_parser.py
@@ -21,7 +21,7 @@ import re
 import sys
 from typing import Union, Any
 
-from utils import DEFAULT_USAGE_MODELS
+from utils import DEFAULT_USAGE_MODELS, DEFAULT_SINGULARITY_IMAGE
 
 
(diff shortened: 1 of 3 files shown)


shreyb:
In draft status until I can run some "integration"-style tests, and maybe even add them to our test suite.

shreyb:
Unit/integration tests, as well as tests described in #330 all pass.

marcmengel (approved):
Looks good!

marcmengel:
Also tested my DAG cases with --line +SingularityImage=  vs --singularity-image, seems to do the right thing, and warns when it's overriding --lines with --singularity-image.

PR #335, opened 2023-03-15
Moved huge amount of site and usage model test cases to data file

shreyb:
Table-driven tests are great, but with this trend, we will see a our test files get huge.  So I moved the test cases for the biggest offender, `test_resolve_site_and_usage_model`, to their own data file.

I did not run entire unit test suite, but I made sure that all the tests in `test_utils_unit.py` pass.


shreyb:
Made same change to singularity image tests

PR #336, opened 2023-03-15
Do not make resource_provides list start with a blank string, 

marcmengel:
…it messes up when we merge them in the DAG code

### DIFF
--- a/lib/get_parser.py
+++ b/lib/get_parser.py
@@ -364,7 +364,7 @@ def get_parser() -> argparse.ArgumentParser:
     parser.add_argument(
         "--resource-provides",
         action="append",
-        default=[""],
+        default=[],
         help="request specific resources by changing condor jdf file. For"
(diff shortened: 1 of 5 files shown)


shreyb (dismissed):
LGTM.  Go ahead and merge into 1.1.4 as we discussed to cut a 1.1.5.

shreyb:
If you need to open a new PR for that to happen, or just want to merge by hand, go ahead (as long as unit tests and any other tests you want pass).

shreyb (changes requested):
This breaks the unit test test_get_parser_all, because that is expecting, like the "--lines" argument, a default [""] list.  I'd split --resource-provides and --lines up for now, and open a ticket to re-investigate to see if we should change how --lines works, or make the unit test cleaner.

shreyb:
Thanks for fixing that unit test.  LGTM.

shreyb (approved):

PR #338, opened 2023-03-16
check .out files for job success, and make sure there are enough 

marcmengel:
So we'll know if our DAG's ran all the nodes or not.


shreyb (changes requested):
Could you please pull in the latest config commit (a6d0e93)?  The one in this PR gets us to 1.1.5, but doesn't seem to include the pycache change you made.

marcmengel:
updated config.

PR #340, opened 2023-03-17
give stack traces if debug/verbose

marcmengel:
This is a second round on error cleanup; print stacktraces if verbose/debug is on.   Have to make the argument list global so we can check verbosity where we call main().

### DIFF
--- a/bin/jobsub_cmd
+++ b/bin/jobsub_cmd
@@ -49,12 +49,12 @@ class StoreGroupinEnvironment(argparse.Action):
         setattr(namespace, self.dest, values)
 
 
-arglist = argparse.Namespace()
+verbose = 0
 
 
 def main() -> None:
     """main line of code, proces args, etc."""
-    global arglist
+    global verbose
     parser = get_parser.get_jobid_parser(add_condor_epilog=True)
     parser.add_argument("-name", help="Set schedd name", default=None)
     parser.add_argument(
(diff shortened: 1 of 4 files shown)


bin/jobsub_cmd
```diff
@@ -49,8 +49,12 @@ class StoreGroupinEnvironment(argparse.Action):
         setattr(namespace, self.dest, values)
 
 
+arglist = argparse.Namespace()
```
shreyb, line 52: So this applies to all four files, but what would you think about _just_ making the verbose value a global?  Since we only need that one variable globally, I'm thinking something along the lines of:

For this line:

```python
verbose = 0
```
Then, down on line 57, it'd be

```python
global verbose
```

And then, after you parse the args on line 71, you could set the global value like:

```python
verbose = args.verbose
```

And down on 240:
```python
if verbose:
    raise
```

There might even be a cleaner way than that, but it's just a thought that occurred to me.

marcmengel: I like that.   Done.

shreyb:
Checked out PR and ran unit tests as well as simply doing:

```
jobsub_submit -G fermilab file:///bad/file
jobsub_submit -G fermilab --debug file:///bad/file
```

The first simply gives us the error that `/bad/file` can't be found, and the second gives the full traceback as intended.

PR #345, opened 2023-03-21
Test fix 0321

marcmengel:
After switching to testing .out files, finding dag.out sambegin.out, etc. so had to adjust a little...


shreyb (approved):
LGTM

PR #356, opened 2023-03-29
Tokens do not work on SL6 containers, so don't set BEARER_TOKEN_FILE

marcmengel:
Several groups using SL6 containers have been running into roadblocks using jobsub_lite;
the main underlying issue is that there is no token support to be had in the underlying
gfal-utilities, etc. on SL6.

### DIFF
--- a/templates/simple/simple.sh
+++ b/templates/simple/simple.sh
@@ -18,6 +18,12 @@ do
 done
 {% endif %}
 
+if grep -q release.6 /etc/system-release
+then
+    : tokens do not work on SL6...
+    unset BEARER_TOKEN_FILE
+else
+
 {% if role is defined and role and role != 'Analysis' %}
 #export BEARER_TOKEN_FILE=$PWD/.condor_creds/{{group}}_{{role | lower}}_{{oauth_handle}}.use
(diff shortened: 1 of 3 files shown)


templates/simple/simple.sh
```diff
@@ -18,6 +18,12 @@ do
 done
 {% endif %}
 
+if grep -q release.6 /etc/system-release
+then
+    : tokens do not work on SL6...
```
shreyb, line 23: Should this be a # (like a comment?)  Or is this some jinja-magic I'm unaware of? 😄

shreyb (changes requested):

templates/simple/simple.sh
marcmengel: Well, ":" is basically "true" - or "pass" in sh scripts.   Initially the then case here was empty; then I  figured out we had added setting BEARER_TOKEN_FILE in the env= of  the .cmd file as well, so 
we have to actively unset BEARER_TOKEN_FILE in that case, so now we don't need a no-op in the
then case anymore, but I left it as a comment. ..

shreyb: Ah - I had no idea! That's nifty.

It LGTM in that case.

shreyb (approved):

Issue #346, opened 2023-03-21
Check all schedds when running commands like jobsub_rm


smitaFnal:
INC000001148876: user submitted a ticket with:
`jobsub_history` with `jobsub_lite` seems to not work with the `--user` flag

1. jobsub_history isn't fully implemented yet.
2. Check all the schedds when running jobsub query commands (rm, history, hold, etc.)

retzkek:
I renamed this since I felt it kinda buried the lede. Should we spin off another issue for rewriting jobsub_history to use landscape (I think that's what we discussed)?

PR #358, opened 2023-03-31
come up with schedd list and some args specific to each schedd.

marcmengel:
This covers the doing things accross multiple schedds part  of #346, but not the jobsub_history part.

### DIFF
--- a/lib/condor.py
+++ b/lib/condor.py
@@ -34,13 +34,13 @@ COLLECTOR_HOST = htcondor.param.get("COLLECTOR_HOST", None)
 
 
 # pylint: disable-next=no-member
-def get_schedd(vargs: Dict[str, Any]) -> classad.ClassAd:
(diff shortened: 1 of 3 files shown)


lib/condor.py
```diff
@@ -34,13 +34,13 @@
 
 
 # pylint: disable-next=no-member
-def get_schedd(vargs: Dict[str, Any]) -> classad.ClassAd:
+def get_schedd_list(vargs: Dict[str, Any]) -> List[classad.ClassAd]:
     """get jobsub* schedd names from collector, pick one."""
```
shreyb, line 38: Could you please update this docstring (and add ones to the two new functions `get_schedd_names` and `get_schedd`) so that in the future, it's very clear which one of these to use?

For the future, do we think this is expansive enough to warrant a `Class` of its own?

shreyb (changes requested):
Aside from the documentation comment I left in `condor.py`, LGTM.

Issue #352, opened 2023-03-23
access other users log files


rlcee:
The old jobsub_fetchlog allowed one user to download another user's log files, if the other user was on the same experiment and the experiment had turned this feature on.  Please implement this feature in jobsub_lite fetchlog. (Suggested in RITM1672696.)

PR #359, opened 2023-03-31
Fetchlog from landscape

retzkek:
Finally it is here! Have `jobsub_fetchlog` download log archives from landscape (using token) instead of from condor. 

* Added `--condor` flag to use `condor_transfer_data` instead
*  Landscape doesn't support zip archives currently. We could add it, but do we really need to support zip? Could also untar and rezip on the client... :raised_eyebrow:
* Set LeaveJobInQueue to False so completed jobs leave queue immediately, but not for DAGMan jobs, since those logs are not currently transferred to dCache/landscape.
* https://github.com/marcmengel/jobsub_lite_config/pull/8

Fixes: #352

### DIFF
--- a/bin/jobsub_fetchlog
+++ b/bin/jobsub_fetchlog
@@ -91,8 +91,16 @@ def fetch_from_condor(
             os.stat(owd)
         except FileNotFoundError:
             os.makedirs(owd, mode=0o750)
-        for f in files:
-            shutil.copy2(os.path.join(iwd, f), owd)  # copy2 tries to preserve metadata
+        try:
+            for f in files:
+                shutil.copy2(
+                    os.path.join(iwd, f), owd
+                )  # copy2 tries to preserve metadata
+        except:
+            print(f"error copying logs to {owd}, leaving them in {iwd}")
+            raise
+        else:
+            shutil.rmtree(iwd)
     else:
         # build archive
         cmd = []
--- a/bin/jobsub_fetchlog
+++ b/bin/jobsub_fetchlog
@@ -139,7 +139,8 @@ def fetch_from_landscape(
 
     if _FETCHLOG_URL_ENV not in os.environ:
         raise Exception(
-            f"{_FETCHLOG_URL_ENV} not set. You may still be able to download with --condor"
+            f"{_FETCHLOG_URL_ENV} not set in the environment. "
+            "You may still be able to download with --condor"
         )
     base_url = os.environ[_FETCHLOG_URL_ENV]
     url = f"{base_url}/job/{jobid}.tar.gz"
--- a/bin/jobsub_fetchlog
+++ b/bin/jobsub_fetchlog
@@ -22,6 +22,9 @@ import sys
 import os
 import subprocess
 import shutil
+from typing import Optional
+import requests  # type: ignore
+from pprint import pprint
 
 if os.environ.get("LD_LIBRARY_PATH", ""):
     del os.environ["LD_LIBRARY_PATH"]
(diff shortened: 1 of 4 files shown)


bin/jobsub_fetchlog
```diff
@@ -51,6 +54,134 @@ from version import print_version, print_support_email
...
+        # If the user wants output in a specific directory, copy files there,
+        # don't build an archive. Old jobsub would get an archive from the
+        # server, upack it into the dest dir, then delete the archive.
+        owd = destdir
+        try:
+            os.stat(owd)
+        except FileNotFoundError:
+            os.makedirs(owd, mode=0o750)
+        for f in files:
+            shutil.copy2(os.path.join(iwd, f), owd)  # copy2 tries to preserve metadata
```
shreyb, line 95: At this point, do we want to remove the IWD?  Or because it's usually in ~/.cache, just let the OS clean it up whenever?

bin/jobsub_fetchlog
```diff
@@ -51,6 +54,134 @@ from version import print_version, print_support_email
...
+def fetch_from_landscape(
+    jobid: str, destdir: Optional[str], archive_format: str, partial: bool
+):
+    # landscape doesn't support zip, does anyone actually use it?
+    if archive_format != "tar":
+        raise Exception(f'unknown/unsupported archive format "{archive_format}"')
+
+    if _FETCHLOG_URL_ENV not in os.environ:
+        raise Exception(
+            f"{_FETCHLOG_URL_ENV} not set. You may still be able to download with --condor"
```
shreyb, line 134: Could you please add "in the environment" or something like that to the first sentence?  Just to make it crystal-clear to the user that they need an env. var `JOBSUB_FETCHLOG_URL`?

bin/jobsub_fetchlog
```diff
@@ -115,63 +252,8 @@ def main():
...
-        p = subprocess.Popen(cmd)
-        if p.wait() != 0:
-            raise Exception(f"error creating archive")
-
-    cleanup({"submitdir": iwd})
-
-    if not transfer_complete:
-        print("Transfer may be incomplete.")
+    fetcher = fetch_from_condor if args.condor else fetch_from_landscape
+    fetcher(args.jobid, args.destdir, args.archive_format, args.partial)
```
shreyb, line 256: 👍 

Love it.

config
shreyb, line 1: Sorry - this was my mistake.  I approved your PR to the  jobsub_lite_config repo as a standard merge commit, so that hash is different than then one here.  Would you mind pulling down the latest jobsub_lite_config commit and updating that here?

If it's too much trouble, I can probably get the hashes to line up to what you have here and force push to jobsub_lite_config.

shreyb (changes requested):
Very nice overall.  I have a few comments, and one request to change one of the helptexts.

shreyb:
Understood.  Do you want to move it back to draft, or keep it as is while you make those changes, if there aren't that many?

FWIW, I was fine with the error handling as you had it, but if you see a deficiency, we'll hold off on merging.

bin/jobsub_fetchlog
```diff
@@ -51,6 +54,134 @@ from version import print_version, print_support_email
...
+):
+    # find where the condor_transfer_data will put the output
+    j = Job(jobid)
+    iwd = j.get_attribute("SUBMIT_Iwd")
+    if verbose:
+        print(f"job output to {iwd}")
+    # make sure it exists, create if not
+    try:
+        os.stat(iwd)
+    except FileNotFoundError:
```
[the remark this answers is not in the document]
retzkek: That's not new though. Did it not work before?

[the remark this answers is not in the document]
shreyb: My impression is that if you submit from one machine, we create the cache dir there, and if you go to another machine to fetch the logs, condor doesn't see the IWD, so condor_transfer_data fails.  This should patch that issue, if I understand correctly.

bin/jobsub_fetchlog
shreyb: LGTM.  Thanks!

bin/jobsub_fetchlog
```diff
@@ -51,6 +54,134 @@ from version import print_version, print_support_email
...
+        try:
+            os.stat(owd)
+        except FileNotFoundError:
+            os.makedirs(owd, mode=0o750)
+        for f in files:
+            shutil.copy2(os.path.join(iwd, f), owd)  # copy2 tries to preserve metadata
+    else:
+        # build archive
+        cmd = []
+        if archive_format == "tar":
```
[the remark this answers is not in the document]
retzkek: In general I like  stronger types, but this is pretty well-scoped, and probably should just be removed as an option.

config
retzkek: No problem, I expected that... 6c12626

retzkek:
I guess I just wanted to include the response text, so pretty small change. Should be GTG now.

bin/jobsub_fetchlog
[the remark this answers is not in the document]
retzkek: Right, but that code has been there since the first version of fetchlog https://github.com/marcmengel/jobsub_lite/commit/d43adaf2592982a8aa852b70c0bb3a35f37ebd89#diff-18845d9de59cabc77f13bd8167605da13f5edc1ae4ab6776ffc44b12c2f94ca4R96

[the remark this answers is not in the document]
shreyb: Ah strange.  Maybe I am getting it mixed up then.

bin/jobsub_fetchlog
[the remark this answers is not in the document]
shreyb: OK.  I leave that choice to you.  As I said above, I'm fine with this as is.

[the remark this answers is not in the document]
retzkek: Then let's stick with it as-is.

bin/jobsub_fetchlog
```diff
@@ -51,6 +54,145 @@ from version import print_version, print_support_email
 
 verbose = 0
 
+# environment variable containing base fetchlog server url
+_FETCHLOG_URL_ENV = "JOBSUB_FETCHLOG_URL"
```
marcmengel, line 58: Personally, I would like to see this defined much closer to where it gets used; like down near line 132.  But don't hold up on merging for this...

marcmengel (approved):
Looks good!  One exceedingly minor comment about moving _FETCHLOG_URL_ENV setting closer to where it is used.

bin/jobsub_fetchlog
shreyb: We can certainly have that discussion.  I tend to do the same as Kevin has here - all constants at the top of the module, but I do see merits to defining it right in the function, especially if it's only going to be used there.  For now, I'll merge, as you said, Marc, but perhaps this discussion can take place in a separate issue or even offline?

shreyb (changes requested):

config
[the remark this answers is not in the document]
retzkek: Done

bin/jobsub_fetchlog
retzkek: Yeah I prefer constants at the top as well, collected together and documented. Or even in their own file.  I've also introduced a new constant type here, the "env var name", which we have hardcoded elsewhere. I'd like to see that more uniform, to make it easier to know what env vars the code uses, and make sure the config aligns.

shreyb (approved):

shreyb:
LGTM.

PR #362, opened 2023-04-04
Update invalid constraint tests so they do not use default jobsub_q output format

shreyb:
In testing the 1.2 release, two integration tests failed because they ran `jobsub_q`, which now prints out the default header, whether or not a job matching the constraint was found.  This broke both tests, because they check the length of the returned output to make sure it's zero.  This change was simply to force jobsub_q to pull down just the job information, since we're testing invalid constraints, and thus there should be no jobs to report on.


marcmengel (approved):
LGTM

PR #363, opened 2023-04-05
Tests cleanup

shreyb:
This PR simply starts to clean up some of our unit tests by doing the following:

1.  Move shared test fixtures into `conftest.py`, which is the [recommended pytest way](https://docs.pytest.org/en/6.2.x/fixture.html#conftest-py-sharing-fixtures-across-multiple-files) of sharing fixtures across multiple test files
2.  Use the fixtures we have like `clear_x509_user_proxy` in older tests where we might have neglected to use the fixture, and are thus duplicating test code.
3. Formally parametrize some of the tests where we run the same test on multiple cases to take advantage of pytests's parametrization functionality (i.e. dynamically generating multiple named tests from a single test so all of the cases run, and we can clearly see which case failed).

Note:  I kept this in draft status so that we don't accidentally merge it before 1.2 is deployed on April 6, 2023; in case we need to make an emergency patch or something like that.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 import os
+import subprocess
 import sys
 
 import pytest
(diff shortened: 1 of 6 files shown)


marcmengel (approved):
Looks good .

shreyb:
Now that 1.2 deployment is complete, moved to review state.  Will merge.

PR #372, opened 2023-04-13
fixes #371, fixes #236

### LINKED ISSUE #236: --mail options don't work
@kherner and @hschellman have both reported that the various --mail* options to `jobsub_submit` do not work.

### LINKED ISSUE #371: JOBSUBJOBSECTION is always 0 for DAG built out of maxconcurrent template
The env var `JOBSUBJOBSECTION` for worker jobs part of DAG built using the ` maxconcurrent` template is always set to `0`, instead of being a progressive index.
An example where this has been observed is the DAG submission `60725507@jobsub01.fnal.gov`.

For DAG built out of `dataset` template  `JOBSUBJOBSECTION` seems to be set properly.

marcmengel:
Just a few missing template lines....

### DIFF
--- a/templates/dag/dag.dag.condor.sub
+++ b/templates/dag/dag.dag.condor.sub
@@ -33,7 +33,7 @@ delegate_job_GSI_credentials_lifetime = 0
 
 {% if generate_email_summary %}
 notify_user = "{{email_to}}"
-notification = "Always"
+notification = Always
 {% endif %}
 
 {% if subgroup is defined and subgroup %}
--- a/templates/dag/dag.dag.condor.sub
+++ b/templates/dag/dag.dag.condor.sub
@@ -32,8 +32,8 @@ delegate_job_GSI_credentials_lifetime = 0
 {{lines|join("\n")}}
 
 {% if generate_email_summary %}
-notify = "{{email_to}}"
-notification = "always"
+notify_user = "{{email_to}}"
+notification = "Always"
 {% endif %}
 
 {% if subgroup is defined and subgroup %}
--- a/templates/dag/dag.dag.condor.sub
+++ b/templates/dag/dag.dag.condor.sub
@@ -31,6 +31,10 @@ delegate_job_GSI_credentials_lifetime = 0
 {%endif%}
 {{lines|join("\n")}}
 
+{% if generate_email_summary %}
+notify= {{email_to}}
+{% endif %}
+
 {% if subgroup is defined and subgroup %}
 +AccountingGroup = "group_{{group}}.{{subgroup}}.{{user}}"
 {% else %}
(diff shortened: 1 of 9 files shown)


shreyb (approved):
LGTM.

shreyb:
Do we want to release this earlier than May 3?  Cut a 1.2.1 that has this, and then a 1.3 with token weakening?

templates/dag/dag.dag.condor.sub
```diff
@@ -31,6 +31,10 @@ delegate_job_GSI_credentials_lifetime = 0
 {%endif%}
 {{lines|join("\n")}}
 
+{% if generate_email_summary %}
+notify= {{email_to}}
```
shreyb, line 35: Actually - hang on, I see other strings are quoted in the submission file.  Does this also need to be quoted?

shreyb (changes requested):

templates/dag/dag.dag.condor.sub
marcmengel: Yes, it probably should be.  Stuff sometimes kind of works when you forget, but...

templates/dag/dag.dag.condor.sub
```diff
@@ -32,7 +32,8 @@ delegate_job_GSI_credentials_lifetime = 0
 {{lines|join("\n")}}
 
 {% if generate_email_summary %}
-notify= {{email_to}}
+notify = "{{email_to}}"
```
shreyb, line 35: Looking at the docs, shouldn't all of these `notify` entries be `notify_user`?

https://htcondor.readthedocs.io/en/latest/man-pages/condor_submit.html

templates/dag/dag.dag.condor.sub
```diff
@@ -32,7 +32,8 @@ delegate_job_GSI_credentials_lifetime = 0
...
+notification = "always"
```
shreyb, line 36: Also, I don't know if condor submission file parsing is case-sensitive, but the same docs as above specify that the accepted value is "Always", not "always", so perhaps this should also be changed.

shreyb (changes requested):
Couple of comments on "notify" vs "notify_user"

templates/dag/dag.dag.condor.sub
marcmengel: Yes!  Thanks..

templates/dag/dag.dag.condor.sub
marcmengel: Changed to upper-case throughout...

shreyb (commented):
LGTM except for the config bit.

shreyb (approved):
Ran all unit tests on this PR, and they all pass. LGTM

PR #373, opened 2023-04-13
closes #281, closes #237

### LINKED ISSUE #237: Feature request: provide feature for bash script to get jobid back from jobsub_submit
From @hschellman:

Is there a way for a bash script to get the jobid back for storage - or have it emailed to me on submit or something? my minerva script does a
submit | grep jobsub which I guess does it. 

@shreyb note:  We'll have to probably discuss with Heidi the parameters around such a feature.

### LINKED ISSUE #281: Add a submit reporting feature
We should have a flag and environment variable (i.e --report-info / JOBSUB_REPORT_INFO) which lets you specify a script to be called with the job-id, command-line arguments, exit status, etc.    This would let us replace our existing poms_jobsub_wrapper infrastructure with a single reporting script, and would let users/experiments have their own script to track jobs they launched, etc.

marcmengel:
adds `--job-info scriptname` option and JOBSUB_EXTRA_JOB_INFO, JOBSUB_EXTRA_LINES, JOBSUB_EXTRA_ENVIRONMENT environment variables to augment the command line argument lists --job-info, --environment , and --lines,

So now you can `export  JOBSUB_EXTRA_ENVIRONMENT=IFDH_DEBUG=1` and all your submissions will get ifdh debugging turned on, without changing your scripts.

But it's really to get rid of the ugliness that is poms_jobsub_wrapper...

### DIFF
--- a/bin/jobsub_submit
+++ b/bin/jobsub_submit
@@ -192,6 +192,11 @@ def transfer_sandbox(src_dir, dest_url):
 verbose = 0
 
 
+def get_env_list(name: str) -> List[str]:
+    """get a split comma separated list of strings from an environment variable"""
+    return [x for x in os.environ.get(name, "").split(",") if x]
+
+
 def main():
     """script mainline:
     - parse args
@@ -210,6 +215,13 @@ def main():
     # so we do that here for backwards compatability
     backslash_escape_layer(sys.argv)
     args = parser.parse_args()
+
+    # Allow environment variables to append to some command lists to get rid of
+    # need for poms_jobsub_wrapper to get in front of us on the path -- we will
+    # just set these and have a job-info script report the job id, etc.
+    args.environment.extend(get_env_list("JOBSUB_EXTRA_ENVIRONMENT"))
+    args.lines.extend(get_env_list("JOBSUB_EXTRA_LINES"))
+    args.job_info.extend(get_env_list("JOBSUB_EXTRA_JOB_INFO"))
 
     verbose = args.verbose
 
--- a/lib/condor.py
+++ b/lib/condor.py
@@ -179,6 +179,12 @@ def submit(
         if m:
             print(f"{hl}Use job id {m.group(1)}.0@{schedd_name} to retrieve output{hl}")
 
+            # call any job_info commands requested with the jobid
+            for ji in vargs["job_info"]:
+                os.system(
+                    f'{ji} {m.group(1)}.0@{schedd_name} "{repr(sys.argv)}" </dev/null'
+                )
+
         return True
     except OSError as e:
         print("Execution failed: ", e)
--- a/lib/get_parser.py
+++ b/lib/get_parser.py
@@ -281,6 +281,12 @@ def get_parser() -> argparse.ArgumentParser:
         " jobs in a DAG",
     )
     parser.add_argument(
+        "--job-info",
+        action="append",
+        default=[],
+        help="script to call with jobid, etc. when job is launched",
+    )
+    parser.add_argument(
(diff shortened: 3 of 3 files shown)


bin/jobsub_submit
```diff
@@ -211,6 +216,13 @@ def main():
     backslash_escape_layer(sys.argv)
     args = parser.parse_args()
 
+    # Allow environment variables to append to some command lists to get rid of
+    # need for poms_jobsub_wrapper to get in front of us on the path -- we will
+    # just set these and have a job-info script report the job id, etc.
+    args.environment.extend(get_env_list("JOBSUB_EXTRA_ENVIRONMENT"))
+    args.lines.extend(get_env_list("JOBSUB_EXTRA_LINES"))
```
shreyb, line 223: Do we want a sanitizer (maybe in utils.py) so that a user can't accidentally throw junk into the submit file?

lib/get_parser.py
```diff
@@ -280,6 +280,12 @@ def get_parser() -> argparse.ArgumentParser:
         help="generate and mail a summary report of completed/failed/removed"
         " jobs in a DAG",
     )
+    parser.add_argument(
+        "--job-info",
+        action="append",
+        default=[],
+        help="script to call with jobid, etc. when job is launched",
```
shreyb, line 287: Is this run when the job is launched or when it is submitted?  From `condor.py`, it looks like the latter.

shreyb (changes requested):

bin/jobsub_submit
marcmengel: I would be okay with that, but what would it sanitize?  I mean, we should probably be doing that to --lines also... would we have a regexp that matches classad definition lines, etc?

lib/get_parser.py
marcmengel: So I reworded it... (I've always used launched and submitted interchangably, but I can see where someone would think launched was when the first job actually started...)

bin/jobsub_submit
marcmengel: I found the Right Answer -- we have Condor check it, with https://htcondor.readthedocs.io/en/latest/apis/python-bindings/api/classad.html#classad.parseOne

shreyb: Love it.

shreyb (commented):
So this is ready except for the config submodule change.

lib/condor.py
```diff
@@ -179,6 +179,12 @@ def submit(
         if m:
             print(f"{hl}Use job id {m.group(1)}.0@{schedd_name} to retrieve output{hl}")
 
+            # call any job_info commands requested with the jobid
+            for ji in vargs["job_info"]:
```
shreyb, line 183: This line makes unit tests fail with a `KeyError`.  We either need to try-except it or have it be something like

```python
job_info = vargs.get("job_info", [])
for ji in job_info
...
```

lib/get_parser.py
shreyb, line 1: Two unit tests for get_parser fail.  Specifically, `test_check_all_test_args` and `test_get_parser_all`.  I think it's because those tests are assuming that the new `--job-info` flag isn't there, or is somehow treating it wrong?

shreyb (changes requested):

lib/get_parser.py
marcmengel: Fixed the unit tests...  Made it so we can't add an option without updating the tests, and now I have to actually do it  ;-)

shreyb:
All unit tests pass now - thanks!

shreyb (approved):

PR #374, opened 2023-04-13
closes #238

### LINKED ISSUE #238: Feature request:  In verbose mode, print out tarball dir structure in job
@hschellman request:  If we pass `--verbose 1`, and a user passes any flag that transfers a file (`-f` or `--tar-file-name`), we should print out the resultant directory structure **IN THE JOB**, so the users can see where they should expect their files to land.  This can be as simple as using environment variables to indicate where the files are, and the top-level directories they can expect with relation to those environment variables.

marcmengel:
Verbose messages about where to find your -f and --tar-file-name files.


bin/jobsub_submit
```diff
@@ -353,6 +353,21 @@ def main():
...
+    if args.verbose:
+        # remind folks where transferred data goes.
+        for f in args.orig_input_file:
+            print(
+                f"File {f}\n ... will be available as: $CONDOR_DIR_INPUT/{os.path.basename(f)}"
+            )
+        i = 0
+        for f in args.orig_tar_file_name:
+            print(
+                f"Contents of {f}\n ... will be avaliable in $INPUT_TAR_DIR_LOCAL{'_'+str(i) if i else ''}"
```
shreyb, line 365: Typo here. "Available"

shreyb (changes requested):
LGTM besides the typo

config
shreyb, line 1: Did you push the config change to github?  I can't see 21e252 in the config repo.

shreyb (changes requested):

config
shreyb: Looking at this again, we can probably keep the config at a352c9 for this PR.

PR #379, opened 2023-04-21
Cleanup integration tests

marcmengel:
Working harder to use the right group / collector to fetch logs.

Dropping the test for repeating schedd's in jobsub_cmd,  as the setup to try to get 2 jobs on 2 schedd's  just doesn't seem to be reliable.

### DIFF
--- a/tests/test_get_parser_unit.py
+++ b/tests/test_get_parser_unit.py
@@ -22,6 +22,20 @@ import get_parser
 import pool
 
 from test_unit import TestUnit
+from test_submit_wait_int import get_collector
+
+
+def set_pool_map():
+    """we need a pool map set to test the --global-pool option"""
+    os.environ["GROUP"] = "dune"
+    os.environ["JOBSUB_POOL_MAP"] = (
+        '{"dune":{"collector":"' + get_collector() + '","onsite":"FNAL_GPGRID"}}'
+    )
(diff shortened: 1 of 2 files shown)


shreyb (changes requested):
The code itself looks good, but two tests fail:

`test_get_parser_all` fails because the `--global-pool dune` flag/arg were added to the `all_test_args` fixture.  The problem is, that won't work unless the `JOBSUB_POOL_MAP` isn't set appropriately, which currently happens only if you've installed the RPM or manually set it.  So that test should either change to exclude the --global-pool flag, and we can have unit tests for pool.py separately, or we should set that environment variable in the test somehow.

The second test that fails is test_launch_lookaround_dune_gp_poolflag, probably for the same reason.

I'm fine keeping this PR out of the 1.3-rc1 release, and then putting it in for 1.3-rc2 in the future.  No need to rush these fixes.

shreyb:
Now that I think about this more, I think the first test should probably exclude --global-pool; and the second test should set the environment variable, since it's an integration test.  Then we can, later on, add pool.py unit tests.

marcmengel:
It seems to me simpler to just set the environment variable in the test... take a look at the latest here.

shreyb (approved):
Thanks for adding the setup.  LGTM.

PR #381, opened 2023-04-24
Fixed key reference errors in condor.py

### LINKED ISSUE #380: Bug:  If jobsub_lite-provided condor_* commands used, we get KeyError on devserver
https://github.com/marcmengel/jobsub_lite/blame/97f2b8d5bf285674b8ed31aab81c9ad61db09da6/lib/condor.py#L42

This line is causing a `KeyError` when we use, for example `condor_submit` as provided by the jobsub_lite libraries.  If using jobsub_lite jobsub_* commands are used, then the `argparse.ArgumentParser` from`get_parser.py` is used to parse the `--devserver` argument, or lack thereof.  If `--devserver` isn't present on the command line, the value is defaulted to `False`.

For the condor_* wrappers we provide, this check is not made, and thus we get the `KeyError`.

This bug was introduced in 1.1.2.

shreyb:
Closes #380 

Fix two KeyError issues that come up if we use condor_submit rather than jobsub_submit.

### DIFF
--- a/lib/condor.py
+++ b/lib/condor.py
@@ -39,7 +39,7 @@ def get_schedd_list(vargs: Dict[str, Any]) -> List[classad.ClassAd]:
     # pylint: disable-next=no-member
     coll = htcondor.Collector(COLLECTOR_HOST)
     # pylint: disable-next=no-member
-    devnot = "" if vargs["devserver"] else "!"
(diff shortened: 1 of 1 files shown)


shreyb:
Note: This PR should go into both master and 1.2-release branches.

PR #382, opened 2023-04-24
Fix 380 for 1.2.2

marcmengel:
Update condor_submit and condor_submit_dag wrappers to pass actual arg-parser args to get_schedd(). 
Also add a test of actually submitting with condor_submit.

### DIFF
--- a/lib/get_parser.py
+++ b/lib/get_parser.py
@@ -157,6 +157,7 @@ def get_base_parser(add_condor_epilog: bool = False) -> argparse.ArgumentParser:
 
 
 def get_submit_parser(add_condor_epilog: bool = False) -> argparse.ArgumentParser:
+    """build the jobsub argument parser for the condor_submit/condor_submit_dag commands and return it"""
     parser = get_base_parser(add_condor_epilog=add_condor_epilog)
     parser.add_argument(
         "--job-info",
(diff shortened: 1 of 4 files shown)


lib/get_parser.py
```diff
@@ -141,6 +141,12 @@ def get_base_parser(add_condor_epilog: bool = False) -> argparse.ArgumentParser:
         action="store_true",
         help="Use jobsubdevgpvm01 etc. to submit",
     )
+    parser.add_argument(
```
shreyb, line 144: Should this actually be here though?  A user, for example, can't run `jobsub_q -G dune --job-info 12345@jobsub01.fnal.gov`, right?

shreyb (changes requested):
Just the 1 thing in get_parser.

lib/get_parser.py
marcmengel: Hm.... well, we could add a third layer; one common to condor_submit, condor_submit_dag and jobsub_submit,
and put the --job-info flag in there.  On the other hand, it is  vacuously correct as is, if jobsub_q **does** call the job-info script for all of the (zero) jobs it suubmits...

lib/get_parser.py
```diff
@@ -156,6 +156,17 @@ def get_base_parser(add_condor_epilog: bool = False) -> argparse.ArgumentParser:
     return parser
 
 
+def get_submit_parser(add_condor_epilog: bool = False) -> argparse.ArgumentParser:
```
shreyb, line 159: Could you please add a quick docstring here explaining that this is the common parser used by all submission binaries (condor_submit, jobsub_submit)?

shreyb (changes requested):
Looks great.  Just the one request for a docstring, and I think this is good to go.

lib/get_parser.py
shreyb: I should have read the file changes before writing that comment.  Thanks for putting in the extra layer of parsers.

PR #391, opened 2023-05-01
Check every RCDS location each retry in job; fail job immediately if we exhaust retries

### LINKED ISSUE #353: Look for RCDS tarball in all repositories each try
During discussion of #324, @vitodb brought up a very good idea for how to handle retries for RCDS tarball discovery on worker nodes when we skip the RCDS check at submission time:

Here:
https://github.com/marcmengel/jobsub_lite/blob/8d7c592bcc686236e68d466f1236fd4a763e96a9/templates/simple/simple.sh#L216

And the corresponding place for --tar-file-name values,

Instead of randomizing our search for the tarballs across all the possible RCDS repositories, and sleeping after each attempt, it is rather inexpensive to just check for the  existence of the directory. Therefore, we should just search for all of the possible locations, and then sleep if none of them work, and try again.

This would look something like, in pseudocode:

```shell
fnamelist=({{fname}})
slp=30
max_retries=40
num_retries=0

while [ num_retries -lt max_retries ]; do 
  num_tries=$(($num_tries + 1))
  fnamelist=( `shuf -e "${fnamelist[@]}"` )  # shuffle our list

  for candidate_fname in ${fnamelist[@]}; do
    if test -f "${candidate_fname}"; then
        copy_file_in()
        break 2 # break out of both loops
    else
      if [[ $num_tries -eq $max_retries ]]; then
           echo "FAILED"
           break
      fi
    fi
  done
  
  sleep $slp  # We didn't find anything, so sleep
done
```

And something similar for the --tar-file-name bit except we're looking for directories, not files.  

It's not like the current way is slow, but this should speed it up, and since it's a relatively inexpensive operation to check the mount points for the existence of files or directories, we might as well take advantage of that.

shreyb:
Closes #353

### DIFF
--- a/templates/simple/simple.sh
+++ b/templates/simple/simple.sh
@@ -223,38 +223,30 @@ export PATH="${PATH}:."
     # Save tfname into a list in case we couldn't get the exact path, like if user skipped the RCDS upload check
     fnamelist=({{fname}})
 
-    # Scale number of tries per input_file possible value.  Maximum 40 (if there's only one possible value
-    # or if we know the path beforehand), minimum 10.  Given these two settings, a single tarball path possibility
-    # can take 5-20 minutes to check
-    len_fnamelist=`echo "${{ '{#' }}fnamelist[@]}"`
+    # Given 40 tries, with a sleep of 30 seconds, we retry for a maximum of around 20 minutes PER item in fnamelist
     num_tries=0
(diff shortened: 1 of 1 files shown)


shreyb:
Going to reopen this against master once I've merged 1.2-series into it.

PR #397, opened 2023-05-03
clear mtime on -f dropbox: tarfile creation.

marcmengel:
This is to address Herb Greenlee's issue that the identical file (a tarball in his case) with a different timestamp keeps getting re-uploaded to RCDS due to the difference in timestamp leading to a different checksum of the tarfile containing that file... 

However, enabling this will also cause a flag day where all -f dropbox: files will get re-uploaded, because they will all (appear to) have a different timestamp (unless they were already at 1 second past the epoch).


shreyb:
We should move this into draft state until we've gotten experiment signoff.  Is that OK with you?

marcmengel:
Moved to draft, pending approval.

shreyb (commented):
This looks good.  Not formally approving yet because we need experiment sign-off.

PR #398, opened 2023-05-04
Dropbox selector as cycle

shreyb:
Realized that a custom-made generator in the `TarfilePublisherHandler` does the same thing as `itertools.cycle`, so replaced that.  Also contains a bugfix where we used to reset the above-mentioned generator every API call rather than with every new CID.  So you could get something like:

```
for cid fermilab/12345

use dropbox_server_1 (fail)
use dropbox_server_2 (fail) 
use dropbox_server_2 (fail)
use dropbox_server_2 (fail)
use dropbox_server_2 (fail)
use dropbox_server_1 (fail)
use dropbox_server_2 (pass)
```

Rather than the intended:

```
for cid fermilab/12345

use dropbox_server_1 (fail)
use dropbox_server_2 (fail) 
use dropbox_server_2 (fail)
use dropbox_server_1 (fail)
use dropbox_server_2 (fail)
use dropbox_server_1 (fail)
use dropbox_server_1 (fail)
use dropbox_server_2 (pass)
```

With this fix, I redesigned it a little bit, so that we might get:

```
for cid fermilab/12345

use dropbox_server_1 (fail)
use dropbox_server_2 (fail) 
use dropbox_server_1 (fail)
use dropbox_server_2 (fail)
use dropbox_server_1 (fail)
use dropbox_server_2 (fail)
use dropbox_server_1 (pass)
```

But for the next cid in the same submission, we might see

```
for cid fermilab/67890

use dropbox_server_2 (fail) 
use dropbox_server_1 (fail)
use dropbox_server_2 (fail)
use dropbox_server_1 (fail)
use dropbox_server_2 (fail)
use dropbox_server_1 (pass)
```

This is a slight redesign, in that we don't shuffle every time we go through the cycle, but that's OK, since the intent is to ensure that we try each dropbox server once in each retry cycle.

### DIFF
--- a/lib/tarfiles.py
+++ b/lib/tarfiles.py
@@ -383,7 +383,7 @@ class TarfilePublisherHandler:
             print(f"Using bearer token located at {self.token} to authenticate to RCDS")
         else:
             print(f"Using X509 proxy located at {self.proxy} to authenticate to RCDS")
-        self.dropbox_servers = tuple(self.dropbox_server_string.split())
+
         self.pubapi_base_url_formatter_full = (
             f"https://{{dropbox_server}}/pubapi/{{endpoint}}"
         )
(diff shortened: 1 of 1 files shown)


marcmengel (approved):
I like  it.   Seeing the addition of the verbose flag though reminds me we should probably have a pass where we convert all the verbose output to the "logging" module, and verbosity would be set by calling into that (not in this request though)

Issue #370, opened 2023-04-10
Make sure all jobsub querying commands (q, rm, hold, release) respect dotted cluster and undotted cluster format


rlcee:
This message output seems confused:
> jobsub_rm --jobid=60475214.@jobsub01.fnal.gov --role=Production

Couldn't find/remove all jobs of user "60475214."

> jobsub_rm --version
jobsub_lite version 1.2

Lisa noted:
condor_q ... 60475214.@jobsub01.fnal.gov ... works but
condor_rm ... 60475214.@jobsub01.fnal.gov ... does not.

Please accept and process this command correctly, or reject it with 
a targeted error message such as "jobid unknown format".

shreyb:
@rlcee - I hope you don't mind: I'm going to change the title of this issue to make it a little more specific.

marcmengel:
So... basically if we stripped trailing dots off of the jobid, we would get consistent results?  Lets see...

marcmengel:
So the change in jobsub_cmd seems to get most of it, now for jobsub_fetchlog...

PR #405, opened 2023-05-11
deal with jobids like 1234.@jobsub01.fnal.gov

marcmengel:
Fixes for #370.  Basically discard the dot from jobids ilke 1234.@jobsubdevgpvm01.fnal.gov


shreyb (approved):
LGTM.

Issue #396, opened 2023-05-03
Uploading files to resilient dCache with -f dropbox:////filename results in error


vitodb:
A jobsub_submit command like this, where it uses `-f dropbox:////filename` , i.e. with 4 "/", with option `--use-pnfs-dropbox`, to deploy the file via resilient, is erroring:
```
jobsub_submit --debug -G mu2e --memory 500MB --disk 2GB --expected-lifetime 1h -f dropbox:///${PWD}/test.tar.gz  --use-pnfs-dropbox file:///usr/bin/sleep 300
Notice: with jobsub_lite, -f dropbox:... does not unpack tarfiles, use --tar_file_name instead
gfal-copy error: 2 (No such file or directory) - Could not stat the source: errno reported by local system call No such file or directory
Traceback (most recent call last):
  File "/opt/jobsub_lite/bin/jobsub_submit", line 402, in <module>
    main()
  File "/opt/jobsub_lite/bin/jobsub_submit", line 273, in main
    do_tarballs(args)
  File "/opt/jobsub_lite/lib/tarfiles.py", line 203, in do_tarballs
    fake_ifdh.cp(pfn, location)
  File "/opt/jobsub_lite/lib/fake_ifdh.py", line 278, in cp
    raise PermissionError(f"Error: Unable to copy {src} to {dest}")
PermissionError: Error: Unable to copy //nashome/v/vito/test.tar.gz to https://fndcadoor.fnal.gov:2880/mu2e/resilient/jobsub_stage/9fa6ff3ef20e57560dd3fa0e6dca8f3665f2f9a11a1cf1ac3eb2842013b94
```
Using the same command without `--use-pnfs-dropbox`  works.

PR #406, opened 2023-05-11
deal with extra slashes in, i.e. , dropbox:////some/place #396

marcmengel:
Small tweak in fake_ifdh.py to keep it from getting confused by extra slashes, or passing them on to gfal-copy and confusing it.   Does not mess with other URL types, etc.

### DIFF
--- a/lib/fake_ifdh.py
+++ b/lib/fake_ifdh.py
@@ -234,6 +234,10 @@ gfal_clean_env = (
 
 
 def fix_pnfs(path: str) -> str:
+
+    if path[0] == "/":
+        path = os.path.realpath(path)
+
     # use nfs4 mount if present
     mountpoint_end = path.find("/", 7)
     if os.path.isdir(path[:mountpoint_end]):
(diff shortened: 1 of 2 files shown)


lib/fake_ifdh.py
```diff
@@ -234,6 +234,10 @@ def generate_proxy_command_verbose_args(cmd_str: str) -> Dict[str, Any]:
 
 
 def fix_pnfs(path: str) -> str:
+
```
shreyb, line 237: Would it be possible to have a unit test or two added to make sure this works as we expect (I mean, it SHOULD, but...)

shreyb (changes requested):
Code itself is good.  Requested unit test.

lib/fake_ifdh.py
marcmengel: On it...

marcmengel: Added.

shreyb: Thanks!

shreyb (approved):
LGTM.

PR #407, opened 2023-05-11
template lines for -L / --log-file

marcmengel:
Add a redirect and in ifdh cp...

### DIFF
--- a/templates/simple/simple.sh
+++ b/templates/simple/simple.sh
@@ -422,8 +422,13 @@ ${JSB_TMP}/ifdh.sh log poms_data=$poms_data
 echo `date` $JOBSUBJOBID BEGIN EXECUTION $JOBSUB_EXE_SCRIPT {{exe_arguments|join(" ")}} >&2
 echo `date` $JOBSUBJOBID BEGIN EXECUTION $JOBSUB_EXE_SCRIPT {{exe_arguments|join(" ")}}
 
-{%if timeout is defined and timeout %} timeout {{timeout}} {%endif%} $JOBSUB_EXE_SCRIPT {{exe_arguments|join(" ")}}
+{%if timeout is defined and timeout %} timeout {{timeout}} {%endif%} $JOBSUB_EXE_SCRIPT {{exe_arguments|join(" ")}} {%if log_file is defined and log_file %}> _joblogfile 2>&1 {% endif %}
 JOB_RET_STATUS=$?
+
+# copy out job log file
+{%if log_file is defined and log_file %}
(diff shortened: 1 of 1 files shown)


marcmengel:
Tested with:
`jobsub_submit -G nova -L /pnfs/nova/scratch/users/mengel/myjob'$CLUSTER.$PROCESS'.log file://lookaround.sh`
seems to work.

templates/simple/simple.sh
```diff
@@ -422,9 +422,14 @@ ${JSB_TMP}/ifdh.sh log poms_data=$poms_data
 echo `date` $JOBSUBJOBID BEGIN EXECUTION $JOBSUB_EXE_SCRIPT {{exe_arguments|join(" ")}} >&2
 echo `date` $JOBSUBJOBID BEGIN EXECUTION $JOBSUB_EXE_SCRIPT {{exe_arguments|join(" ")}}
 
-{%if timeout is defined and timeout %} timeout {{timeout}} {%endif%} $JOBSUB_EXE_SCRIPT {{exe_arguments|join(" ")}}
+{%if timeout is defined and timeout %} timeout {{timeout}} {%endif%} $JOBSUB_EXE_SCRIPT {{exe_arguments|join(" ")}} {%if log_file is defined and log_file %}> _joblogfile 2>&1 {% endif %}
```
shreyb, line 425: Pardon my ignorance - but will we still get stdout and stderr in the standard job log locations and thus in Landscape?  I think so...

marcmengel: No.  Well, at least not stdout and stderr of running the job executable.   It all goes in the log and gets copied out.
This is the behavior jobsub_client had.(  jobsub_client also had a flag to compress the log ; but no-one has asked
for that one. )

marcmengel: (It also dodges any truncation of the the stdout/stderr that our stuff does, BTW)

shreyb: So basically, if someone uses the `-L` option, they will not get stdout and stderr in the normal landscape locations, and thus the Landscape/POMS sandbox links will send them to a place with empty files?  And then `jobsub_fetchlog` would also similarly pull down "log files" that were empty?

Is there some magic with `tee` we could use to preserve the log in both places?  If it's too difficult, I'm fine with giving the explanation of "we replicated jobsub_client behavior".

marcmengel: Well.... I could tee it into a logfile and send it all to stdout as well; but we lose the stdout/stderr split if we do that.

shreyb: Ah - then let's forget it.  Let's just go ahead and call it good.

PR #408, opened 2023-05-15
Made args_for_schedd into a defaultdict because we always expect list values

shreyb:
Just a quick change I thought about after reviewing #405.  Since we always expect a list in `args_for_schedds`, rather than do a check for schedd membership in that dict, we can simply make it a default dict which will add the key automatically if it doesn't exist.

Tested this doesn't break anything (and still solves #370, like #405 did) with the following:

```
 ~/jobsub_lite/bin/jobsub_submit -G fermilab --debug -N 20 --devserver file:///usr/bin/sleep 300
 ~/jobsub_lite/bin/jobsub_q -G fermilab --devserver --jobid=11420.0@jobsubdevgpvm01.fnal.gov
 ~/jobsub_lite/bin/jobsub_rm -G fermilab --devserver --debug --jobid=11420.@jobsubdevgpvm01.fnal.gov
```

### DIFF
--- a/bin/jobsub_cmd
+++ b/bin/jobsub_cmd
@@ -23,6 +23,7 @@
 # pylint: disable=wrong-import-position,wrong-import-order,import-error
 
 import argparse
+from collections import defaultdict
 import os
 import sys
 import re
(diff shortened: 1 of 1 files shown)


marcmengel (approved):
Nice.  Easier to read.

PR #415, opened 2023-05-16
More pnfs fun

marcmengel:
Problem Lisa reported with chmod failing re-adding pnfs persistent file...

### DIFF
--- a/lib/fake_ifdh.py
+++ b/lib/fake_ifdh.py
@@ -253,6 +253,8 @@ def chmod(dest: str, mode: int) -> None:
         os.chmod(dest, mode)
     except FileNotFoundError as e:
         pass
+    except PermissionError as e:
+        pass
 
 
 def mkdir_p(dest: str) -> None:


lib/fake_ifdh.py
```diff
@@ -253,6 +253,8 @@ def chmod(dest: str, mode: int) -> None:
         os.chmod(dest, mode)
     except FileNotFoundError as e:
         pass
+    except PermissionError as e:
```
shreyb, line 256: Nitpicky - do we ever expect to handle these two exceptions differently?  If not, we should probably make this `except (FileNotFoundError, PermissionError) as e`.  But if we think we might expand the exception handling, leave it as is.

Either way, LGTM

marcmengel: I think we might want to do something more interesting here later; ignoring it entirely is a broad brush... but for now I'm not thinking of any specific cases we want to handle differently...

shreyb:
@goodenou :  Hopefully this should fix the issue you raised on Slack.

Issue #417, opened 2023-05-17
Multiple submits with same tarfile name from same directory clash


marcmengel:
The current code makes a filtered tarfile (with permissions corrected) in the current working directory, with
a name based on the name of the tarfile being submitted.
If two submissions are being made at the same time, with the same tarfile basename, in the same directory,
this causes a clash and the tarfile gets overwritten and removed; this means that one job can
get the other job's tarfile, and that one of the submissions can fail from the tarfile "disappearing"  because
the other jobsub_submit removed it.

shreyb:
@hgreenlee originally reported this issue.

shreyb:
Note:  This also sounds like an old ticket we got from mu2e:  ServiceNow ticket INC000001150554.

PR #403, opened 2023-05-10
Small bug fixes

marcmengel:
adding exit(1) to the except clauses in most scripts to fix #402, Additions to the requirements= line to fix  #401

PR #420, opened 2023-05-17
put pid in tarfile names we create to avoid clashes

marcmengel:
This to avoid the problems of #417, or similarly, problems with two job submissions doing
`--tar-file-name tardir:///same/path`  from the same directory at the same time, we put the
process-id in the generated tarfile names.

### DIFF
--- a/lib/tarfiles.py
+++ b/lib/tarfiles.py
@@ -73,7 +73,7 @@ def check_we_can_write() -> None:
 
 def tarchmod(tfn: str) -> str:
     """copy a tarfile to a compressed tarfile changing modes of contents to 755"""
-    ofn = os.path.basename(f"{tfn}.tbz2")
+    ofn = os.path.basename(f"{tfn}{os.getpid()}.tbz2")
     check_we_can_write()
     try:
         with tarfile_mod.open(tfn, "r|*") as fin, tarfile_mod.open(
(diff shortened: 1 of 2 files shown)


shreyb (commented):
The code changes look good to me.  

I submitted a test job that I've used before to check RCDS upload functionality:

```
$ jobsub_submit -G fermilab  --tar-file-name tardir:///home/sbhat/TestJobs/TestDir --tar-file-name  dropbox:///home/sbhat/FileUploadTests/testfiles.tar --onsite file:///home/sbhat/FileUploadTests/tar-file-name.sh 2
```

The job id of that submission is `61821641.0@jobsub01.fnal.gov`.  As long as that runs fine, I'll be happy to call this good to go.

shreyb:
Note:  all tarfile unit tests pass as well, so that's good.

shreyb:
This job isn't running due to a bug introduced in #403.  See #422 for details.  If you're OK with it, @marcmengel , I'd like for us to resolve #422 first before merging this (I'll work on it first thing tomorrow morning - I think it's pretty easy), because without #422 resolved, we can't even submit jobs to test.

shreyb:
Resubmitted test job with PR checked out plus manually adding the change of #423.

Job id is `61829142.0@jobsub01.fnal.gov`

If job runs successfully, the PR is ready to be merged.

shreyb (approved):
Looks good and all tests pass.

PR #423, opened 2023-05-18
make blacklist unset check work..

### LINKED ISSUE #422: Issue with --blacklist implementation
Note that **THIS ISSUE IS A BLOCKER FOR 1.3.3**.

In #403, a bug was introduced that makes jobs not able to run at all.  While testing #420, I submitted a job, and noticed that it didn't run quickly.  The submit command was:

```
$ jobsub_submit -G fermilab  --tar-file-name tardir:///home/sbhat/TestJobs/TestDir --tar-file-name  dropbox:///home/sbhat/FileUploadTests/testfiles.tar --onsite file:///home/sbhat/FileUploadTests/tar-file-name.sh 2
```

Upon looking at `condor_q <jobid> -bet`, I saw the following in the job requirements:

```
      ((isUndefined(target.GLIDEIN_Site) == false) &&
        ( !stringListIMember(target.GLIDEIN_Site,my.Blacklist_Sites)))) && (TARGET.Arch == "X86_64") 
```

Since I didn't specify a `--blacklist` value, this was surprising.  Looking at the template for simple.cmd, we see this clause, introduced in #403:  
```
{%if blacklist is defined and blacklist != '' %} && ((isUndefined(target.GLIDEIN_Site) == FALSE) && (!stringListIMember(target.GLIDEIN_Site,my.Blacklist_Sites))){%endif%}
```

The intent of #403 is to implement a similar check to the sites check in the job requirements:

```
{%if site is defined and site != '' %} && ((isUndefined(target.GLIDEIN_Site) == FALSE) && (stringListIMember(target.GLIDEIN_Site,my.DESIRED_Sites))){%endif%}
```

However, the `--sites` argument defaults to `""`, whereas `--blacklist` currently has no default, so it gets set by argparse to `None` if no value is given.

I think it'll be easier if we simply have `--blacklist` default to `""`, as long as that passes the unit tests introduced in #404, and a couple of simple submission tests.

marcmengel:
This changes the default for --blacklist to "" (empty string) instead of None to fix #422

### DIFF
--- a/lib/get_parser.py
+++ b/lib/get_parser.py
@@ -199,6 +199,7 @@ def get_parser() -> argparse.ArgumentParser:
     parser.add_argument(
         "--blacklist",
         help="ensure that jobs do not land at these (comma-separated) sites",
+        default="",
     )
     parser.add_argument("-r", help="Experiment release version")
     parser.add_argument("-i", help="Experiment release dir")


marcmengel:
Tested submissions ran both with _and without_ --blacklist this time.

shreyb (approved):

PR #424, opened 2023-05-19
Add --schedd-for-testing flag

### LINKED ISSUE #421: --jobsub-schedd-for-testing flag
We should have a hidden (`help=argparse.SUPPRESS` in the argument declaration) argument, like `--jobsub-schedd-for-testing`, that allows us to select a jobsub schedd.

The idea is that this flag would not be advertised, and ONLY intended for testing.  We don't want to _hide_ it per se, but we don't want to encourage its use among end users.

shreyb:
Closes #421 

I also changed the `condor_status -schedd` code to use a jinja template, since the logic with this added feature was getting a bit tougher to parse.

Tested with no schedd specified, a couple of our real schedds, and a fake schedd.  Jobs were properly directed to the requested schedd, or the submission rejected in the case of the fake schedd.  All jobs (`sleep 300`) succeeded.

Also, ran all unit tests, which pass.

### DIFF
--- a/lib/get_parser.py
+++ b/lib/get_parser.py
@@ -21,9 +21,10 @@ import re
 import sys
 from typing import Union, Any, List
 
-from utils import DEFAULT_USAGE_MODELS, DEFAULT_SINGULARITY_IMAGE
 import pool
 from skip_checks import SupportedSkipChecks
+from utils import DEFAULT_USAGE_MODELS, DEFAULT_SINGULARITY_IMAGE
+from condor import get_schedd_names
 
 
 def verify_executable_starts_with_file_colon(s: str) -> str:
@@ -89,6 +90,26 @@ class VerifyAndAddSkipCheck(argparse.Action):
             setattr(namespace, self.dest, checks_to_skip)
             new_arg_to_set = f"skip_check_{values}"
             setattr(namespace, new_arg_to_set, True)
+
+
+class CheckIfValidSchedd(argparse.Action):
+    """Action to check if the tester has requested a valid schedd to submit to"""
+
+    def __call__(
+        self,
+        parser: argparse.ArgumentParser,
+        namespace: argparse.Namespace,
+        values: Any,
+        option_string: Union[None, str] = None,
+    ) -> None:
+        group = os.environ.get("JOBSUB_GROUP", os.environ.get("GROUP", None))
+        vargs = {"group": group} if group is not None else {}
+        valid_schedds = get_schedd_names(vargs)
+        if values not in valid_schedds:
+            raise TypeError(
+                f"Invalid schedd specified: {values}.  Valid choices are {valid_schedds}"
+            )
+        setattr(namespace, self.dest, values)
 
 
 def get_base_parser(add_condor_epilog: bool = False) -> argparse.ArgumentParser:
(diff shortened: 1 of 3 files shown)


lib/get_parser.py
```diff
@@ -427,6 +448,12 @@ def get_parser() -> argparse.ArgumentParser:
         ' +DESIRED_CVMFS="OSG" to the job classad attributes and'
         " '&&(CVMFS==\"OSG\")' to the job requirements",
     )
+    parser.add_argument(
```
marcmengel, line 451: Just a thought: should we put this in the base parser, so we can set the schedd for jobsub_q, etc.?

marcmengel (approved):
I like it, but wanted to consider the comment about maybe putting it in the base parser...

lib/get_parser.py
shreyb: Sure - I can move it there, and then we can implement it in a separate issue?

shreyb: Sorry - that was unclear.  I can move the argument to the base parser, and then we can implement the functionality for jobsub_q in a separate issue?

shreyb: So this actually already exists, by passing the `-name` flag.  There's just a bug in the code right now that ignores the `-name` flag:

```
$ jobsub_q -G fermilab  -name jobsub01.fnal.gov sbhat
JOBSUBJOBID                             OWNER           SUBMITTED     RUNTIME   ST PRIO   SIZE  COMMAND
688176.0@jobsub04.fnal.gov              sbhat           05/16 16:15   0+00:19:53 C    0    0.0 dagman_wrapper.sh -p 0 -f -l . -Lockfi
688179.0@jobsub04.fnal.gov              sbhat           05/16 16:15   0+00:00:22 C    0    0.0 dagman_wrapper.sh -p 0 -f -l . -Lockfi
503632.0@jobsub05.fnal.gov              sbhat           05/11 16:10   0+00:01:07 C    0    0.0 dagman_wrapper.sh -p 0 -f -l . -Lockfi
677046.0@jobsub05.fnal.gov              sbhat           05/19 19:50   0+00:18:43 C    0    0.0 dagman_wrapper.sh -p 0 -f -l . -Lockfi
61563115.0@jobsub01.fnal.gov            sbhat           05/10 16:46   0+00:00:08 C    0    0.0 simple.sh 
61859314.0@jobsub01.fnal.gov            sbhat           05/19 19:50   0+00:00:22 C    0    0.0 dagman_wrapper.sh -p 0 -f -l . -Lockfi
68343522.0@jobsub02.fnal.gov            sbhat           05/11 16:10   0+00:17:53 C    0    0.0 dagman_wrapper.sh -p 0 -f -l . -Lockfi
```

It's a quick fix, and with the next commit, it works right, with the `-name` flag and not the `--schedd-for-testing` flag:

```
$ jobsub_q -G fermilab  -name jobsub01.fnal.gov sbhat
JOBSUBJOBID                             OWNER           SUBMITTED     RUNTIME   ST PRIO   SIZE  COMMAND
61563115.0@jobsub01.fnal.gov            sbhat           05/10 16:46   0+00:00:08 C    0    0.0 simple.sh 
61859314.0@jobsub01.fnal.gov            sbhat           05/19 19:50   0+00:00:22 C    0    0.0 dagman_wrapper.sh -p 0 -f -l . -Lockfi
```

shreyb: Actually - I'll make that a separate PR.  I'll move the `--schedd-for-testing` flag to the base parser, and if you approve, we'll merge it.  I'll open another PR for the bugfix for the issue immediately above, just so we can track the progress more easily and it doesn't all become one huge merge commit eventually.

shreyb:
@marcmengel - I see that you already approved the PR, but does the last commit address your comment?

Aside:  I opened #425 to allow for `-name` to be used in `jobsub_q`, `jobsub_release`, etc.

marcmengel:
Looks good.

marcmengel:
Yes, that last change to move it to the base parser was what I was looking for; that makes it work for the condor_submit wrapper, etc.

PR #433, opened 2023-06-07
adding totals to jobsub_q default output

marcmengel:
... by piping to a simple total script.

### DIFF
--- a/bin/jobsub_cmd
+++ b/bin/jobsub_cmd
@@ -224,16 +224,16 @@ def main() -> None:
                 ]
             )
 
-            # pipe our output through jobsub_totals
-            totalsf = os.popen("jobsub_totals", "w")
-            os.close(1)
-            os.dup2(totalsf.fileno(), 1)
-
             print(
                 "JOBSUBJOBID                             OWNER       \tSUBMITTED     RUNTIME"
                 "   ST PRIO   SIZE  COMMAND"
             )
             sys.stdout.flush()
+
+            # pipe our remaining output through sort (by date) and jobsub_totals
+            totalsf = os.popen("sort -k 3,4 | jobsub_totals", "w")
+            os.close(1)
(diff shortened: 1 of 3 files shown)


marcmengel:
... and "sort".

marcmengel:
So there is another commit pending here.;. not showing up in the web interface yet for some reason...

bin/jobsub_cmd
```diff
@@ -230,6 +230,11 @@ def main() -> None:
             )
             sys.stdout.flush()
 
+            # pipe our remaining output through sort (by date) and jobsub_totals
+            totalsf = os.popen("sort -k 3,4 | jobsub_totals", "w")
```
shreyb, line 234: Since we assume that `jobsub_cmd` is in the same directory as `jobsub_totals`, is it reasonable to hardcode the relative path here?  I only ask because my testing of this doesn't work because without installing the RPM, `jobsub_cmd` can't find `jobsub_totals` (since it's looking in `/opt/jobsub_lite/bin`).  I'm thinking something like:

```python
jobsub_totals_path = os.path.join(os.path.dirname(__file__), "jobsub_totals")
# ....
totalsf = os.popen(f"sort -k 3,4 | {jobsub_totals_path}", "w")
```

Something like that, roughly?  What do you think?

shreyb (commented):
So my only concern in testing this is that for VOs with a huge amount of submitted jobs (I checked DUNE), it takes a lot longer (~10s vs. ~3s) for the output to start to appear since we're piping everything through sort by default.  Do you think we should make the sorting bit an argument the user can pass?  I worry that someone will think something is wrong when their jobsub_q doesn't return within a couple of seconds.

marcmengel:
Well.. it is slower to give the first response; but it is still faster than old-jobsub jobsub_q...

shreyb (approved):

Issue #333, opened 2023-03-14
Requested enhancements to --dataset-definition behavior


marcmengel:
At the meeting with uboone this morning, Herb mentioned two features we could add to the -dataset-definition DAG behavior
that would let their scripts stop making their own DAGs:
1) allow setting percent threshold for staging check in startup job
2) allow giving extra datasets to check in startup job for alternate workflows

We should consider whether adding these to jobsub_lite would be of general utility.

PR #434, opened 2023-06-08
Dataset definition enhancements 

marcmengel:
Implementation of #333. 

Adds 2 flags --dd-percentage and --dd-extra-dataset which set:
* percentage of dataset files to wait for in start job
* extra dataset(s)  to wait for staging in start job (option can be repeated)


marcmengel:
Reviewers should hide whitespace in the diffs, as a block is indented...

marcmengel:
Okay, so I discussed some more with Greenlee... the divisor flag I added isn't what he's looking for, it's really a percentage of the files in the dataset -- so rather than dividing the prestage limit; he wants to do something like
`TOTAL_FILES=$(( TOTAL_FILES * {{dd_percentage}} / 100 ))`
so marking this a draft 'till I get that changed...

shreyb (commented):
This looks good - but one question.  Is the new integration test a full test of the functionality?  That is, me or another dev who doesn't know much about SAM/prestaging could just run that test and be confident that if the jobs run and exit 0, that the advertised functionality works?

marcmengel:
Well, we could test that the percentage and the extra-dataset show up in the sambegin.sh...  
But I'm really not sure how to automate getting a dataset 30% staged for it to *really* test, 
at least not consistently -- I mean if someone else staged it completely, I can't go kick it out of
DCache....  Maybe if we {mocked? proxied with filtering?} the samweb server to report the project
as 30% cached and pointed it at that?

shreyb:
If you're OK with it, would it be too much work to implement the checking of sambegin.sh in a test?  At least that way we know we're giving the batch system the right instructions.

marcmengel:
> If you're OK with it, would it be too much work to implement the checking of sambegin.sh in a test? At least that way we know we're giving the batch system the right instructions.

Added a test_render_dd_flags in test_jobsub_submit_unit.py to make sure if we render files for datset-definition, and have dd_percentage and dd_extra_datasset values, they show up in the sambegin.sh

shreyb (approved):
Looks great.  Thanks!

PR #435, opened 2023-06-09
Store schedd ads in memory via global variable; give more user-friendly error message if MAX_PROCS_PER_SUBMISSION exceeded

### LINKED ISSUE #367: Make error message for submitting too many jobs in a cluster from Condor more user-friendly
Right now, if we try to submit a cluster with too many jobs, we see the following error message:

```
$ jobsub_submit -G fermilab -N 100000 file:///home/kretzke/testinfo.sh
Attempting to get token from https://htvaultprod.fnal.gov:8200 ... failed
<omitted for conciseness>
ERROR: Failed to create proc
Number of submitted jobs would exceed MAX_JOBS_PER_SUBMISSION
Submitting job(s)................................................................ (lots more dots)
=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-

Error: condor_submit exited with failed status code 1
=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
```

This is fine, but I would think most users would want to know _what_ that limit is.  Modify/replace this error message so that we actually tell them what MAX_JOBS_PER_SUBMISSION is if they get this error.

shreyb:
Closes #367 

Along with fixing #367, I realized that for this, and possibly other cases, we're making multiple calls to the collector to get schedd info, where we really only need that once, since we don't retain state between executable runs.  For that reason, I also changed `get_schedd_list` to write to a global var, `__schedd_ads`, which is simply a dict of `schedd_name: schedd_ad`.  `get_schedd_list` then checks if this dictionary is populated before querying the collector.  If it does need to query the collector (and this can be forced via a new argument to the function), it will populate `__schedd_ads` as well.

Other functions inside the `condor.py` module can then also use this global variable to query schedd ads, which is what I use here to get the `Jobsub_Max_Procs_Per_Submission` attribute from the applicable schedd.

I generally try to stay away from globals, but I think in this case, it makes sense.

### DIFF
--- a/lib/condor.py
+++ b/lib/condor.py
@@ -36,6 +36,10 @@ random.seed()
 # pylint: disable-next=no-member
 COLLECTOR_HOST = htcondor.param.get("COLLECTOR_HOST", None)
 
+# Global dict to hold onto schedd ads so we only have to query schedd once
+# This should ONLY be changed by get_schedd_list
+__schedd_ads: Dict[str, classad.ClassAd] = {}
+
 
 @contextmanager
 def submit_vt(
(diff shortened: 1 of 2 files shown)


marcmengel (approved):
Looks good.

PR #436, opened 2023-06-09
some unit tests worked standalone, but were tripping over other tests…

marcmengel:
… actions


tests/test_condor_unit.py
shreyb, line 1: Great call.  Thanks for catching that.

shreyb (approved):
Looks good.

PR #444, opened 2023-06-20
utils.cleanup() was failing if cache directory was already removed

marcmengel:
This was printing an error in jobsub_fetchlog --condor.

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -38,6 +38,8 @@ DEFAULT_SINGULARITY_IMAGE = (
 
 
 def cleandir(d: str) -> None:
+    if not os.path.exists(d):
+        return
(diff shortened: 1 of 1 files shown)


lib/utils.py
```diff
@@ -46,7 +48,7 @@ def cleandir(d: str) -> None:
 
 def cleanup(varg: Dict[str, Any]) -> None:
     """cleanup submit directory etc."""
-    os.chdir(f'{varg["submitdir"]}/..')
+    os.chdir(os.path.dirname(f'{varg["submitdir"]}'))
```
shreyb, line 51: I like this just in general.

shreyb (approved):
LGTM

PR #445, opened 2023-06-21
fix the fix

marcmengel:
note the cmd change 10 lines earlier...
 cmd = cmd.replace("jobsub_", "condor_")

thus we need to check for cmd == "condor_q", not cmd == "jobsub_q"... sigh


shreyb:
Oh weird - I checked the original fix and it seemed to work.  Not sure why....  But this looks good.

shreyb (approved):

shreyb (changes requested):
The change is good, but could you please update the config submodule to the right commit?

Issue #440, opened 2023-06-14
Ignore --maxConcurrent if --maxConcurrent m  -N  n and n < m 


marcmengel:
Some users are including --maxConcurrent  10000 in all submissions, which is launching DAGs all the time, 
even when doing -N 1 jobs...  We can basically ignore the -maxConcurrent in this case, which will reduce 
the number of DAGs and make debugging those jobs when they go wrong easier.

shreyb:
Do this before #438, since it's easier and fixes the underlying user issues.

PR #447, opened 2023-06-29
implements #440

marcmengel:
Just a little logic check: if the --maxConcurrent value is greater than -N for how many jobs, just drop it. 
I'm doing this in jobsub_submit, just after we

### DIFF
--- a/bin/jobsub_submit
+++ b/bin/jobsub_submit
@@ -274,6 +274,13 @@ def main():
 
     varg = vars(args)
 
+    if args.maxConcurrent and int(args.maxConcurrent) > args.N:
+        if args.verbose:
+            sys.stderr.write(
+                f"Note: ignoring --maxConcurrent {args.maxConcurrent} for {args.N} jobs\n"
+            )
+        args.maxConcurrent = None
+
     proxy, token = get_creds(varg)
 
     if args.verbose:
--- a/bin/jobsub_submit
+++ b/bin/jobsub_submit
@@ -274,7 +274,7 @@ def main():
 
     varg = vars(args)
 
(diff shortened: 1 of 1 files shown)


shreyb (commented):
Should the `varg` assignment come after this change so that `varg` picks up the change?  Or is that deliberately excluded?

shreyb:
This doesn't concern just this PR, but I'm starting to feel like there's too much going on in the `jobsub_submit` main.  Maybe we should consider exporting some of this functionality to functions, other modules, etc so that our `submit` logic is very clear.  I'm especially thinking of all the places where we manipulate args.

This doesn't have to be addressed here, if you agree, but if you do agree, I'll open an issue to clean that up in the future.

marcmengel:
> Should the `varg` assignment come after this change so that `varg` picks up the change? Or is that deliberately excluded?

Yes!  Ack!  Good catch...

marcmengel:
> This doesn't concern just this PR, but I'm starting to feel like there's too much going on in the `jobsub_submit` main. Maybe we should consider exporting some of this functionality to functions, other modules, etc so that our `submit` logic is very clear. I'm especially thinking of all the places where we manipulate args.
> 
> This doesn't have to be addressed here, if you agree, but if you do agree, I'll open an issue to clean that up in the future.

But yes, I agree we should probably factor some of that out into another source file...  certainly all the functions defined in jobsub_submit besides main, and then probably the blocks of render code  in if args.dag:...elif args.dataset_dag:...else: could be factored out into subroutines in another module.

shreyb (approved):
LGTM

PR #449, opened 2023-06-30
Bl a o cklist

marcmengel:
Changing --blacklist to --blocklist, throughout, with --blacklist as a backwards-compatability spelling.


lib/utils.py
```diff
@@ -674,14 +674,14 @@ def __init__(self, site: str, usage_model: str):
 
 class SiteAndBlacklistConflictError(Exception):
```
shreyb, line 675: Should we change the name here as well?  (and at line 655 where this is raised)

tests/test_utils_unit.py
```diff
@@ -299,32 +299,32 @@ def test_resolve_singularity_image(self, singularity_test_case):
...
+        "site_blocklist_test_case",
+        site_blocklist_test_data(good=False),
         ids=create_id_for_test_case,
     )
-    def test_check_site_and_blacklist_bad(self, site_blacklist_test_case):
+    def test_check_site_and_blocklist_bad(self, site_blocklist_test_case):
         """Test to make sure that a given comma-separated site list string
-        and blacklist string are handled correctly"""
+        and blocklist string are handled correctly"""
         with pytest.raises(utils.SiteAndBlacklistConflictError):
```
shreyb, line 326: Same comment here - should we change this to `SiteAndBlocklistConflictError`?

tests/test_utils_unit.py
```diff
@@ -111,14 +111,14 @@ def singularity_test_data():
     return [SingularityTestCase(**test_json) for test_json in tests_json]
 
 
-def site_blacklist_test_data(good=True):
-    """Pull in site/blacklist test data from data file and return a list of
+def site_blocklist_test_data(good=True):
+    """Pull in site/blocklist test data from data file and return a list of
     test cases"""
     SiteAndBlacklistTestData = namedtuple(
```
shreyb, line 117: We should probably change this to `SiteAndBlocklistTestData`, as well as the return value of this function

shreyb (commented):
This looks good, functionally.  There are a couple of places where "blacklist" wasn't changed to "blocklist", in the names of objects.  I think it's safe to change those names too, since those objects are pretty localized.

tests/test_utils_unit.py
marcmengel: So yes, getting those...
There's also a +Blacklist_Sites classad, if we change that do we have to change monitoring, etc.?

shreyb: Yeah - I think we should keep those as is, because that's the infrastructure side, and may have unforeseen downstream consequences.

shreyb (approved):
LGTM

PR #450, opened 2023-06-30
Clean env vs dash e

marcmengel:
Computing list of environment variables in utils.py, list now includes LC_CYTPE, but is also checked against -e/--environment variables.   Added a shorter "smoke" test category to the test suite that submits one job and checks it.

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -305,17 +305,35 @@ def set_extras_n_fix_units(
     fix_unit(args, "expected_lifetime", timtable, -1, "smhd", -1)
     fix_unit(args, "timeout", timtable, -1, "smhd", -1)
     newe = []
+    envnames = set()  # set of environment variable names with -e for below
     for e in args["environment"]:
         pos = e.find("=")
         if pos < 0:
+            envnames.add(e)
             v = os.environ.get(e, None)
             if not v:
                 raise RuntimeError(
                     f"--environment {e} was given but no value was in the environment"
                 )
             e = f"{e}={v}"
+        else:
+            envnames.add(e[0:pos])
+
         newe.append(e)
+
     args["environment"] = newe
+
+    #
+    # build list of environment variables for wrapper script to clear:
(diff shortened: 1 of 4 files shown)


lib/utils.py
```diff
@@ -305,17 +305,35 @@ def set_extras_n_fix_units(
...
+
         newe.append(e)
+
     args["environment"] = newe
+
+    #
+    # build list of environment variables for wrapper script to clear:
+    # -- this is our default list, below, minus anything passed in a -e/--environment argument
+    #
+    full_clean_env_list = ["LC_CTYPE", "CPATH", "LIBRARY_PATH"]
```
shreyb, line 330: This is absolutely fine as is.  

That being said, it's be nice if we could alert the user if any of their variables specified in `-e` was also in this list, and thus we will NOT be clearing it.  I think this could be easily accomplished if `full_clean_env_list` were another set, and then you could have something like

```python
clean_env_list = full_clean_env_list.difference(envnames)
```

(and cast it to a list if jinja doesn't like sets).

And then we could inform the user that the environment variables given in `full_clean_env_list.intersection(envnames)` will not be cleared out, as they are specified in the `-e` argument.

What are your thoughts about that?

shreyb (commented):
This looks good, but I had one small suggestion.

lib/utils.py
marcmengel: Well... I think alerting the user is maybe Too Much Information for most users, and will generate "why am I getting this error message?" tickets... we could add a comment in the wrapper script maybe, so people who are looking at that level of detail can see it?

marcmengel: Do you like this latest?

shreyb (approved):

lib/utils.py
shreyb: That's a good point.  The way you made this, we can always revisit this if folks get confused.  We should make sure to document it in the environment variables section for the wrapper script docs that -e now overrides the clearing-out list.

PR #452, opened 2023-07-10
Get rid of tarfile size memory limits by streaming upload

marcmengel:
So now we open the tarfile twice; once to checksum it, and the second time to get a fresh filehandle to pass as the data for requests.post for the upload.  This way we do not read the whole tarfile into memory, and cause memory size issues on large tarfiles.

### DIFF
--- a/lib/tarfiles.py
+++ b/lib/tarfiles.py
@@ -506,12 +506,11 @@ class TarfilePublisherHandler:
         url = self.pubapi_cid_url_formatter.format(endpoint="publish")
         if self.verbose:
             print(f"Calling URL {url}")
-        if self.token:
-            with open(tarfilename, "rb") as tarfile:
+
+        with open(tarfilename, "rb") as tarfile:
+            if self.token:
                 return requests.post(url, auth=TokenAuth(self.token), data=tarfile)
-        res = requests.post(url, cert=(self.proxy, self.proxy), data=tarfile)
-        tarfile.close()
-        return res
+            return requests.post(url, cert=(self.proxy, self.proxy), data=tarfile)
 
     @cid_operation
     @pubapi_operation
--- a/lib/tarfiles.py
+++ b/lib/tarfiles.py
@@ -124,7 +124,7 @@ def tar_up(directory: str, excludes: str, file: str = ".") -> str:
     return tarfile
 
 
-def slurp_file(fname: str) -> Tuple[str, BinaryIO]:
+def checksum_file(fname: str) -> str:
     """pull in a tarfile while computing its hash"""
     h = hashlib.sha256()
     with open(fname, "rb") as f:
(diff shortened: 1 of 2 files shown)


lib/tarfiles.py
```diff
@@ -124,19 +124,17 @@ def tar_up(directory: str, excludes: str, file: str = ".") -> str:
...
     with open(fname, "rb") as f:
         tff = f.read(4096)
         h.update(tff)
-        tfl.append(tff)
         while tff:
             tff = f.read(4096)
             h.update(tff)
-            tfl.append(tff)
-    return h.hexdigest(), bytes().join(tfl)
+    tf = open(fname, "rb")
```
shreyb, line 136: I'm OK with opening the file twice - I think that's a clean solution.

What really worries me is the readability of opening a file here and closing it in a completely different place.  Is there any reason we can't modify this function to just return the checksum, and then in line 315, pass the filename in as a string to `publisher.publish()`?

Since we're opening the file twice anyway, we could then wrap the `requests.post` call in a file open/close context manager; something like 

```python
with open(tarfile, 'rb') as tarfile_handler:
    res = requests.post(url, auth=TokenAuth(self.token), data=tarfile_handler)
    return res
```

What do you think?

shreyb (changes requested):

lib/tarfiles.py
marcmengel: I kind of don't like the close-it-elsewhere bit either.. And of course, then this routine becomes *just* a checksummer,
as we're passing the name into it... I'll try that.

marcmengel: There, I think that's better.

lib/tarfiles.py
```diff
@@ -124,19 +124,16 @@ def tar_up(directory: str, excludes: str, file: str = ".") -> str:
     return tarfile
 
 
-def slurp_file(fname: str) -> Tuple[str, bytes]:
+def checksum_file(fname: str) -> str:
```
shreyb, line 127: Ah - good call changing the name of the function.

lib/tarfiles.py
```diff
@@ -510,8 +507,11 @@ def publish(self, tarfile: str) -> requests.Response:
         if self.verbose:
             print(f"Calling URL {url}")
         if self.token:
-            return requests.post(url, auth=TokenAuth(self.token), data=tarfile)
-        return requests.post(url, cert=(self.proxy, self.proxy), data=tarfile)
+            with open(tarfilename, "rb") as tarfile:
+                return requests.post(url, auth=TokenAuth(self.token), data=tarfile)
+        res = requests.post(url, cert=(self.proxy, self.proxy), data=tarfile)
```
shreyb, line 512: If we don't have a token, then I believe this and the next line will fail with a `NameError` since `tarfile` isn't defined in the local scope.   Unless github is messing with the view, I believe the intended scoping (no `close()` necessary) is:

```python
with open(tarfilename, "rb") as tarfile:
    if self.token:
        return requests.post(url, auth=TokenAuth(self.token), data=tarfile)  # Token
    return requests.post(url, cert=(self.proxy, self.proxy), data=tarfile)  # Proxy
```

Is that right?

shreyb (changes requested):

lib/tarfiles.py
marcmengel: Yes.   Fixed. Of course, if we can't get  a token, we won't get this far... but better to have it right.

shreyb: Thanks!

Issue #451, opened 2023-07-04
jobsub_submit refactoring


shreyb:
We should probably factor some of that out into another source file...  certainly all the  functions defined in jobsub_submit besides main, and then probably the blocks of     render code  in if args.dag:...elif args.dataset_dag:...else: could be factored out into subroutines in another module.

_Originally posted by @marcmengel in https://github.com/fermitools/jobsub_lite/issues/447#issuecomment-1614938052_

PR #453, opened 2023-07-11
Refactor submit

marcmengel:
Implementation of #451.    Created a new lib/submit_support.py with code that was already subroutines in jobsub_submit, and some more code split out to make the mainline more manageable.   Boundary between this code and utils.py  routines is unclear, should maybe have just made them utils1.py and utils2.py or something.

Note that this pull request attempts to make no changes to actual code; just *moves* it to another file and 
adds a few `def function_name(args, schedd_name): ` lines.   So wherever you see a block of code replaced by a subroutine call, you should later find that exact code in the body of that subroutine.

This probably leaves jobsub_submit with some imports it doesn't need any more, if folks are looking for more small improvements.

### DIFF
--- a/bin/jobsub_submit
+++ b/bin/jobsub_submit
@@ -204,7 +204,7 @@ def main():
     elif args.maxConcurrent:
         jobsub_submit_maxconcurrent(varg, schedd_name)
     else:
-        jobsub_submit_simple(varg)
+        jobsub_submit_simple(varg, schedd_name)
 
     if args.verbose:
         # remind folks where transferred data goes.
--- a/bin/jobsub_submit
+++ b/bin/jobsub_submit
@@ -44,7 +44,7 @@ PREFIX = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 sys.path.append(os.path.join(PREFIX, "lib"))
 from packages import pkg_find
 
-pkg_find("jinja")
+pkg_find("jinja")  # type: ignore
 import jinja2 as jinja
 
 #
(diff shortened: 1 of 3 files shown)


shreyb (commented):
This is VERY nice.  Given the huge scope of "changes", I do want to run this through the unit tests and a few other tests before approving it.

marcmengel:
I am a little worried that this will cause merge clashes with the tracing pull request... Maybe I'll try and merge them in a branch somewhere and see...   Actually not so bad.  It causes a merge conflict  in jobsub_submit, but it's pretty easy to clean up.

retzkek:
> This probably leaves jobsub_submit with some imports it doesn't need any more, if folks are looking for more small improvements.

Does our linting not check for unused imports? Maybe it should, I'm not a huge fan of leaving them around.

shreyb:
> > This probably leaves jobsub_submit with some imports it doesn't need any more, if folks are looking for more small improvements.
> 
> Does our linting not check for unused imports? Maybe it should, I'm not a huge fan of leaving them around.

Mypy does ignore imports:

https://github.com/fermitools/jobsub_lite/blob/master/.pre-commit-config.yaml#L17C40-L17C65

However, in our Pylint does not:

https://github.com/fermitools/jobsub_lite/blob/master/.pylintrc#L428

We can investigate this in a separate issue to make sure we've got this covered.  My recollection is because of how we set up the relative module importing, mypy was always complaining about imports, so we told it to ignore the unused imports.

retzkek:
Thanks, that's fine we can deal with any unused imports in that issue.

PR #456, opened 2023-07-14
Adding Fermitools license file

marcmengel:


shreyb (changes requested):
Is that filename supposed to be COPYRIGHT rather than COPYING?

shreyb:
And should that be "LICENSE"?  I can't remember what the Fermilab rules are off the top of my head.

marcmengel:
So I checked my email logs; Aaron recommended an Apache 2.0 license for this; so yes, that one generally goes in a LICENSE file.  So I dropped the other one, and added this one.

marcmengel:
Hang on, re-read the notes...

shreyb (approved):

Issue #270, opened 2023-02-08
make proxy requirement optional for job submission


vitodb:
The CI has a use case where the proxy provided is shorter that 167h and there is no kerberos ticket available to get a longer proxy.
I'd like to have an option to opt out for the proxy requirement.
If the submission happens with this option enabled, a currently available proxy, even if with short lifetime, is sent to the job, it would be fine.

shreyb:
Started work on this.  Underlying library code done.  Just need to make changes to callers so they can support getting proxy, token, or both.

shreyb:
@marcmengel - I'm hoping to have this done today or tomorrow, since I'm on vacation next week.

shreyb:
Pretty much done.  All unit tests pass.  I want to do the following "basic" tests before opening a PR for it:

- [ ] Basic submit (done [2472079.0@jobsub05.fnal.gov](https://landscape.fnal.gov/lens/view/job/2472079.0@jobsub05.fnal.gov/r/summary))
- [ ] Basic Submit, Token-only
- [ ] Basic submit, proxy-only
- [ ] Basic submit, invalid auth method (DONE)
- [ ] Basic submit with file transfer
- [ ] Basic submit with file transfer, token-only
- [ ] Basic submit with file transfer, proxy-only

shreyb:
Some small KeyErrors with the proxy-only submission.  I got it to work, but I want to reexamine where we rely on BEARER_TOKEN_FILE being set.

shreyb:
Fixed all errors, and all unit tests and submissions as detailed above work.  Will wait until jobs run to make sure that correct credentials land in job.

Bash history to detail my submits in tests:

```
 ~/jobsub_lite/bin/jobsub_submit -G fermilab --debug file:///usr/bin/sleep 300
 ~/jobsub_lite/bin/jobsub_submit -G fermilab --debug --auth-methods token file:///usr/bin/sleep 300
 ~/jobsub_lite/bin/jobsub_submit -G fermilab --debug --auth-methods proxy file:///usr/bin/sleep 300
 ~/jobsub_lite/bin/jobsub_submit -G fermilab --debug --auth-methods proxy,token file:///usr/bin/sleep 300
 ~/jobsub_lite/bin/jobsub_submit -G fermilab --debug --auth-methods proxy,token,willfail file:///usr/bin/sleep 300
 cd ~/FileUploadTests/
[...]

PR #459, opened 2023-07-21
Make obtaining proxy optional in job submission

shreyb:
This is a massive PR, so please take your time going through it.  It adds a few things:

1. A couple of constants where we configure which auth methods are defaulted to, and which are required by our infrastructure (`creds.DEFAULT_AUTH_METHODS`, `creds.REQUIRED_AUTH_METHODS`).
2.  A new object, `creds.CredentialSet`, that stores the paths to the various supported credential types.  Its instantiation automatically stores the paths to the credentials in the correct environment variables (e.g. `BEARER_TOKEN_FILE`.
3.  A new argument, `--auth-methods`, that can be passed in.  So `--auth-methods token` requests only a token, `--auth-methods "proxy,token"` requests both, etc.  There are safeguards in place at both the parser level and the underlying creds code (see next item) to prevent invalid auth methods from being requested.  The order of preference is:
  1. `--auth-methods` flag
  2. `JOBSUB_AUTH_METHODS` environment variable
  3. `creds.DEFAULT_AUTH_METHODS` list
4. `creds.get_creds()` now will check the requested auth methods to make sure that they include the `REQUIRED_AUTH_METHODS`, and will return a `CredentialSet`.  Since most callers don't actually do anything with the return value of `get_creds()` and simply use it to ensure credentials are refreshed and in the environment, I didn't need to change much on the caller end.
5. Unit tests to test all of this.

The tests I ran aside from the unit tests are listed in #270, so please feel free to take a look.  Let me know if you have any questions.

### DIFF
--- a/bin/condor_submit
+++ b/bin/condor_submit
@@ -41,7 +41,7 @@ sys.path.append(os.path.join(PREFIX, "lib"))
 # import our local parts
 #
 from condor import get_schedd, submit
-from creds import get_creds
+from creds import get_creds, print_cred_paths_from_credset
 import get_parser
 
 # pylint: disable-next=no-member
@@ -123,10 +123,9 @@ def main():
         )
         return
 
-    proxy, token = get_creds(vars(args))
+    cred_set = get_creds(vars(args))
     if args.verbose:
-        print(f"proxy is : {proxy}")
-        print(f"token is : {token}")
+        print_cred_paths_from_credset(cred_set)
 
     #   SUBMIT
 
--- a/bin/condor_submit_dag
+++ b/bin/condor_submit_dag
@@ -37,7 +37,7 @@ sys.path.append(os.path.join(PREFIX, "lib"))
 # import our local parts
 #
 from condor import get_schedd, submit_dag
-from creds import get_creds
+from creds import get_creds, print_cred_paths_from_credset
 import get_parser
 
 
@@ -66,11 +66,9 @@ def main():
     if arglist.verbose:
         cmd_args.append("-debug")
 
-    proxy, token = get_creds(arglist)
-
-    # put in environment for condor libs to use
-    os.environ["X509_USER_PROXY"] = proxy
-    os.environ["BEARER_TOKEN_FILE"] = token
+    cred_set = get_creds(vars(arglist))
+    if arglist.verbose:
+        print_cred_paths_from_credset(cred_set)
 
     if os.environ.get("GROUP", None) is None:
         raise SystemExit(f"{sys.argv[0]} needs -G group or $GROUP in the environment.")
--- a/bin/jobsub_cmd
+++ b/bin/jobsub_cmd
@@ -35,7 +35,7 @@ if os.environ.get("LD_LIBRARY_PATH", ""):
 PREFIX = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 sys.path.append(os.path.join(PREFIX, "lib"))
 
-import fake_ifdh
+import creds
 import get_parser
 import condor
 from version import print_version, print_support_email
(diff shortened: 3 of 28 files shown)


shreyb:
@marcmengel's comments from Slack:

```
So the only thing I see you might want to change, is maybe we should pass the credential set into set_extras_and_fix_units, and  make the bit at:

https://github.com/fermitools/jobsub_lite/blob/master/lib/utils.py#L134

be a loop over the credentials(?)   Also, it looks like maybe the get_client_dn  call before that should maybe be skipped if there isn't a proxy...
```

```
It looks like get_client_dn will whine if there is really no proxy around...
```

shreyb:
@marcmengel - I think the last two commits address the issues you raised.  We still whine, but if no proxy was requested, we don't even try to get the clientdn.

marcmengel (approved):
Looks excellent.   I especially like all the tests.

shreyb:
Since I'm out for the next week, I see no rush to merge this before next week. However, if you guys do want to merge #131, assuming Kevin is OK with it, go for it and I'll rebase.  I'd be interested to know if this is easy to use for @vitodb , for example.

Did DAGs look good with this PR, Marc?  I submitted one and it worked, but I'm very inexperienced when it comes to DAGs and their intricacies.

marcmengel:
Okay, so while things look good, something in test_creds_unit.py breaks test_dagnabbit_unit.py tests  (and others) if  you run for example pytest -m unit.   pytest test_dagnabbit_unit.py standalone runs fine...
Looks like we need to reset something?!?

marcmengel:
So.. the last test in test_creds_unit.py sets the proxy location to "proxylocation", which sets `os.environ["X509_USER_PROXY"] = "proxylocation"`, which confuses everything after it.

So we need a `del os.environ["X509_USER_PROXY"]` at the end of the last test in test_creds_unit.py...

shreyb:
I think I'll add that to every test that doesn't have the `clear_x509_user_proxy` fixture.

marcmengel:
okay, now we're down to a few more:
in TestUtilsUnit.test_get_client_dn_valid_proxy_provided
...
decode_token.sh could not successfully extract the expiration time from token file /tmp/pytest-of-mengel/pytest-25/test_tarfile_publisher_cid_ope0/fake_location. Please open a ticket to Distributed Computing Support if you need further assistance.
...
>           return int(exp_time) - time.time() > 60
E           ValueError: invalid literal for int() with base 10: ''
in our needs_credentials fixture

marcmengel:
And then another one:
E                   PermissionError: Error: Unable to create /pnfs/fermilab/users/mengel/jobsub_stage/8a77b29e6ab2bcc88c7b7e711c656f0eacadd14e92c9a8c2bdddeb9bf2406257/tmpa3vrky8816541.tgz16541.tbz2

shreyb:
It's weird - I always run these individually in VSCode.  Maybe I should start running them through pytest as well.  Either way, I just fixed both of those, I think.  New commits incoming.

shreyb:
These pass a `pytest -m unit` run, btw.

marcmengel:
Trying submission tests now... will have results in a few hours.

marcmengel:
Ah.. but 4 submission tests failed;
 File "/home/mengel/jobsub_lite/templates/dag/dag.dag.condor.sub", line 22, in template

    {{% if clientdn is defined and clientdn %}}+JobsubClientDN="{{clientdn}}"{{% endif %}}

jinja2.exceptions.TemplateSyntaxError: unexpected '%'

That should be` {%if`  not` {{%if`

marcmengel:
E               subprocess.CalledProcessError: Command '['jobsub_fetchlog', '--group', 'fermilab', '--jobid', '13149.0@jobsubdevgpvm01.fnal.gov', '--destdir', '/tmp/tmpo_5y8lz8']' returned non-zero exit status 1.

/usr/lib64/python3.6/subprocess.py:438: CalledProcessError
----------------------------------------------------- Captured stderr call -----------------------------------------------------
jobsub_fetchlog only supports token authentication.  Please either omit the --auth-methods flag or make sure tokens is included in the value of that flag

Even if you *add* `--auth-methods token`, it still says thils ..  ahh.. it's spelled "tokens" and not "token" in the check -- jobsub_fetcblog line 245.

shreyb:
Agh - I thought I had fixed that `{{}}` issue.  Sorry about that.  Also fixed `jobsub_fetchlog`.  Pushing commits now.

marcmengel:
Rerunning full test suite on this branch; but I'm pretty sure its going to run clean now..

marcmengel:
Yay!  All tests pass!

PR #470, opened 2023-08-25
tarfile file-size checks

marcmengel:
Checking the file size while we convert tarfiles, and raising an exception for files > 1G, which would fail later when and RCDS upload is attempted.

### DIFF
--- a/lib/tarfiles.py
+++ b/lib/tarfiles.py
@@ -47,6 +47,8 @@ except ValueError:
     )
     raise
 
+RCDS_MAX_FILE_SIZE = 1073741824
+
 
 class TokenAuth(AuthBase):  # type: ignore
     # auth class for token authentication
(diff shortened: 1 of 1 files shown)


lib/tarfiles.py
```diff
@@ -88,6 +92,11 @@ def tarchmod(tfn: str) -> str:
                 else:
                     st = fin.extractfile(ti)
                 ti.mode = ti.mode | 0o755
+                if ti.size > 1073741824:
```
shreyb, line 95: Would it be possible to maybe make this either a package-level constant or a variable at the beginning of this function, to improve readability?  Something like 

```python
RCDS_tarfile_file_size_limit = 1073741824
...
if ti.size > RCDS_tarfile_file_size_limit
...
```

(that's a horrible variable name, but you get the idea).

marcmengel: Done.

shreyb (approved):
LGTM

PR #472, opened 2023-08-29
Adding a 'jobsub' command to list available commands

### LINKED ISSUE #471: Add a "jobsub -h" option to list all jobsub commands
A request from Heidi Schellman in RITM1845157

Users often don't know which jobsub commands exist.  If there were a jobsub -h (or even jobsub ) that just listed the jobsub commands, that would be very helpful.  Bonus points for having it point to a url for the real documentation.

marcmengel:
It's pretty much more documentation than code, but here it is.

Closes #471

### DIFF
--- a/bin/decode_token.sh
+++ b/bin/decode_token.sh
@@ -44,7 +44,7 @@ usage() {
 
 case x$1 in
 x-e) extractfilt="get_field $2"; shift; shift;;
-x-h) usage; exit 0;;
+x-h|x--help) usage; exit 0;;
 x)   usage; exit 1;;
 *)   extractfilt="cat";;
 esac
--- a/bin/jobsub
+++ b/bin/jobsub
@@ -38,5 +38,29 @@ jobsub_lite also wraps the following condor commands to accept '-G group':
     All condor commands will print usage with -help and
     "man condor_..." will give a manual page, or see:
     https://htcondor.readthedocs.io/en/latest/man-pages/index.html
+
+jobsub_lite commands are affected by many environment variables, particularly:
+
+    BEARER_TOKEN_FILE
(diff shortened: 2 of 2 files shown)


shreyb:
Converted to draft so we can discuss this at the next Developers meeting on Aug 30, 2023.

retzkek (changes requested):
I'd like to see one-line descriptions of the commands here. This may also be a good place for a list of environment variables that affect how jobsub operates.

Also I'd like us to consider (not necessarily implement here) a refactor of jobsub to the common subcommand pattern, e.g. `jobsub_submit ...` becomes `jobsub submit ...` (with a compatibility alias), which would make sure that the list of (sub-)commands and their descriptions don't stray from what's actually implemented, and would help ensure uniformity between them. Maybe we include the wrapped condor commands too, e.g. `jobsub condor_q ...`.

This is important to discuss now as it could impact how this "quick fix" script works, to make sure that it won't become a future breaking change.

marcmengel:
So, I've added a brief description to the jobsub commands,and minimal command line flag information.
and (some) environment variables; this ramps us up to two and a half screenfuls of description...
I am concerned this is a little too much information for an overview...

marcmengel:
So, looking at the master "jobsub" command idea again... If we took the existing
main() functions in the the 4 main jobsub scripts
jobsub_cmd
jobsub_fetchlog
jobsub_history
jobsub_submit
(the rest are all symlinks...) 
and split them into 2 parts -- 1 that adds arguments to the parser, and 1 that
implements the script logic, we could then have standalone scripts which 
would be pretty much:

if __name__ == '__main__':
       parser = argparser.ArgumentParser
       parsr = this_command_get_parser(parser=parser)

PR #478, opened 2023-09-27
let schedd-for-testing override downtime checks, etc.

marcmengel:


lib/get_parser.py
```diff
@@ -237,7 +237,7 @@ def get_base_parser(add_condor_epilog: bool = False) -> argparse.ArgumentParser:
     parser.add_argument(
         "--schedd-for-testing",  # Non-advertised option for testers to direct jobs to certain schedds
         type=str,
-        action=CheckIfValidSchedd,
+        # action=CheckIfValidSchedd,
```
shreyb, line 240: Why does this need to be commented out?  On my reading of the code, it seems like `CheckIfValidSchedd` serves here as a check that the requested `schedd_for_testing` is registered to the configured collector.  Is that not the case?

shreyb (changes requested):

lib/get_parser.py
marcmengel: It limits you to the active schedds, so in the case that came up yesterday, it wouldn't let you pick jobsub03 because it had the in maintenance flag turned on, whereas we want to submit there to test it before we open it up...

marcmengel: We could instead turn it back on , and have it call get_schedd_list with a flag that gets *all* the schedd's, rather than just the active ones(?)...

marcmengel: But for the moment, folks can use this branch to test jobsub03...

marcmengel: Okay, now the check is back on, but with the *full* list of schedd's to choose from, not just the ones that are not in downtime, or available for our group, or even have IsJobsubLite...

shreyb (changes requested):
LGTM, but could you please back the config change out of this PR?  Since you've got another PR open for it, it'll create a different hash when I merge that commit, which will throw this branch off needlessly.

shreyb:
I wish github would allow for fast-forward merges, because then what you did with config would be totally OK...

Issue #477, opened 2023-09-26
In tarfiles.tarfile_in_dropbox, we don't get credentials for the pnfs dropbox case


shreyb:
Reproduce the issue:

```
$ jobsub_submit -G gm2 --use-pnfs-dropbox  --tar-file-name tardir:///home/sbhat/FileUploadTests -n --verbose 3 file:///usr/bin/sleep 300
./
./testfile2
./testfile3
./tar-dash-f-mix.sh
./dash-f-dropbox.sh
./testfiles.tar
./tar-file-name.sh
./testfile
./dash-f.sh
gfal-mkdir error: 13 (Permission denied) - HTTP 401 : Authentication Error
Traceback (most recent call last):
  File "/opt/jobsub_lite/bin/jobsub_submit", line 233, in <module>
    main()
  File "/opt/jobsub_lite/lib/tracing.py", line 161, in wrapper
    res = func(*args, **kwargs)
  File "/opt/jobsub_lite/bin/jobsub_submit", line 153, in main
    do_tarballs(args)
  File "/opt/jobsub_lite/lib/tracing.py", line 161, in wrapper
    res = func(*args, **kwargs)
  File "/opt/jobsub_lite/lib/tarfiles.py", line 266, in do_tarballs
    path = tarfile_in_dropbox(args, tfn[8:])
  File "/opt/jobsub_lite/lib/tarfiles.py", line 364, in tarfile_in_dropbox
    fake_ifdh.mkdir_p(os.path.dirname(location))
  File "/opt/jobsub_lite/lib/fake_ifdh.py", line 276, in mkdir_p
    raise PermissionError(f"Error: Unable to make directory {dest}")
PermissionError: Error: Unable to make directory https://fndcadoor.fnal.gov:2880/GM2/resilient/jobsub_stage/6fbfd6cc7a960f82dca01aa4dadcef2e03819a7d4a387664c49db71a0bedabb1
```

This happens because in `tarfiles.tarfile_in_dropbox` - https://github.com/fermitools/jobsub_lite/blob/master/lib/tarfiles.py#L301 - we only get the credentials in the CVMFS case (like this line: https://github.com/fermitools/jobsub_lite/blob/master/lib/tarfiles.py#L316).  In the PNFS case (`--use-pnfs-dropbox`), we don't do the same, so job submission will fail due to inability to authenticate.  

[...]

Issue #476, opened 2023-09-13
Currently, putting --need-storage-modify, etc. on dagman job lines has no effect...


marcmengel:
While we may be able to make that work in a move to local-universe dagman jobs in the future, at the moment
having --need-storage-modify and --need-scope flags on individual entries in the DAG has no effect.

So if they differ from the ones the overall DAG is submitted with, we should print a warning. 
This will mean comparing two lists of options / removing duplicates (and we should  be removing
duplicates there when we merge the options anyhow, so as not to make the scope excessively long
with repeated directories, etc.)   This would be around:
https://github.com/fermitools/jobsub_lite/blob/master/lib/dagnabbit.py#L201
which is in the midst of a chunk of code that is merging the options from the jobsub_submit(_dag) command
line with the options from the dagnabbit submit line.   

And later, if we *do* make the --need-storage-modify and --need-scope flags work on individual DAG
stages, we  should *still*  check that the DAG ones are a subset of the overall ones, and warn if they're
not, as the submission of that stage will otherwise fail once we make it work. 

And while we're in there, that section of dagnabbit.py could probably stand to be factored out into several
subroutines, with descriptive names, and tests, it's a bit monolithic already, and this code will only make
that worse...

Issue #473, opened 2023-08-31
TRACEPARENT formatting in v1.4 preventing environment variables from being passed to lens


LTrestka:
jobsub_lite is not formatting the "TRACEPARENT" environment variable correctly according to condor docs, this causes lens to fail at parsing.

All variables passed into the submission make it to the jobsub logs, and to kibana, but nothing in the list that comes after and including TRACEPARENT can be queried in lens.

ex:
`Environment = "POMS4_CAMPAIGN_STAGE_NAME=dd_sample_campaign_stage_1 GRID_USER=ltrestka TRACEPARENT=\"00-e3131325f54a321cdb504f049bfcac76-3976795c459dbf45-01\" IFDH_DEBUG=1 EXPERIMENT=hypot POMS4_CAMPAIGN_STAGE_ID=2602 CLUSTER=71832400 JOBSUBJOBID=71832400.0@jobsub02.fnal.gov POMS4_CAMPAIGN_NAME=dd_sample_campaign_1 "`
- In this example , only POMS4_CAMPAIGN_STAGE_NAME and GRID_USER would make it to lens.
- Note: I removed some variables to keep it simple, full submission can be seen [here](https://fifemon.fnal.gov/kibana/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:now-24h,mode:quick,to:now))&_a=(columns:!(_source),index:'fifebatch-history-*',interval:auto,query:(language:lucene,query:'JobsubJobId:71832400.0@jobsub02.fnal.gov'),sort:!('@timestamp',desc)))

Per the doc:
>Put double quote marks around the entire argument string. This distinguishes the new syntax from the old. The old syntax does not have double quote marks around it. Any literal double quote marks within the string must be escaped by repeating the double quote mark.
https://htcondor.readthedocs.io/en/latest/man-pages/condor_submit.html (ctrl-f "enviroment")

retzkek:
From what I can tell it's condor that's formatting the environment incorrectly - we actually specify it in the submit script with the "old" format. https://github.com/fermitools/jobsub_lite/blob/master/templates/simple/simple.cmd#L18

Issue #466, opened 2023-08-08
Makefile enhancement to make it AL9-compatible


shreyb:
Add a:

OSR=$(shell /bin/sh -c ' . /etc/os-release; echo $$VERSION_ID | sed -e s/\\..*//')

and some:

ifeq ($OSR,7) ... endif

_Originally posted by @marcmengel in https://github.com/fermitools/jobsub_lite/pull/464#pullrequestreview-1567656092_

This is to make sure we don't try to build the Tracing RPM on AL9 when we run `make`.

PR #479, opened 2023-09-28
Bug fixes

marcmengel:
Fixes for #477 #476 #473 #466 #460 #458, and for swallowing stderr of condor_submit.

### DIFF
--- a/templates/simple/simple.cmd
+++ b/templates/simple/simple.cmd
@@ -12,10 +12,10 @@ log                = {{filebase}}.log
 {%if not ( is_dag is defined and is_dag ) %}
 JOBSUBJOBSECTION=$(Process)
 {%endif%}
-TRACEPARENT={{traceparent}}
+TRACEPARENT="{{traceparent}}"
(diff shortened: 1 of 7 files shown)


marcmengel:
Tests are clean now.

config
shreyb, line 1: Like #478 , could you please back out this config update so that to make the eventual update at release time a bit cleaner?

templates/simple/simple.cmd
shreyb, line 1: I like this solution.

shreyb (changes requested):
Looks great, except for the config submodule bit.  Thanks for fixing all of these!

config
marcmengel: Done.

marcmengel:
I will note that the swallowing of condor_submit's stderr outupt was probably the cause of several tickets, where we got failures with no other error message visible -- condor_submit said something on stderr, and we threw it away :-1:

shreyb:
Oh REALLY.... that's good to know.

shreyb (approved):

PR #480, opened 2023-09-29
Use scitokens library rather than decode_token.sh

marcmengel:
It turns out to be pretty straightforward, with one caveat: more exceptions from down below.  
Pretty much if anything went wrong with decode-token.sh, you got an empty string back
with ValueError being the only exception you got back; the new one has a half dozen or
so exceptions, and validates tokens for expiration, etc. all the time.   Minimal code changes
for it, although the `except Exception ` bit in checkToken may be overly broad.


lib/token_mods.py
```diff
@@ -55,9 +56,12 @@ def use_token_copy(tokenfile: str) -> str:
 def get_token_scope(tokenfilename: str) -> List[str]:
     """get the list of scopes from our token file"""
 
-    with os.popen(f"decode_token.sh -e scope {tokenfilename}", "r") as sf:
-        data = sf.read()
-        scopelist = data.strip().strip('"').split(" ")
+    with open(tokenfilename) as f:
```
shreyb, line 59: I think this is good for now, but in the future, time permitting, we should look at refactoring the 

```python
with open(tokenfilename) as f:
    ...
    token = dict(....)
```

bit as a single function that can be shared.  I'm not sure where the correct place for such a function would be though...

lib/fake_ifdh.py
```diff
@@ -94,21 +102,19 @@ def checkToken(tokenfile: str) -> bool:
...
+            token_encoded = f.read(4096).strip()
+            token = dict(
+                scitokens.SciToken.deserialize(token_encoded, insecure=True).claims()
+            )
+            exp_time = token["exp"]
+            add_event(f"expiration: {exp_time}")
+            return int(exp_time) - time.time() > 60
+    except Exception:
+        # anything else, just say its bad
+        return False
```
retzkek, line 117: Is this the right place to catch the exception? Seems the caller is in a better position to handle it. Like in `getToken()`, with the first call we don't care why it's invalid since we'll get a new one, but with the second call we should pass along why the newly-acquired token is not valid.

lib/fake_ifdh.py
```diff
@@ -94,21 +102,19 @@ def checkToken(tokenfile: str) -> bool:
...
+        exp_time = None
+        with open(tokenfile) as f:
+            token_encoded = f.read(4096).strip()
+            token = dict(
+                scitokens.SciToken.deserialize(token_encoded, insecure=True).claims()
+            )
+            exp_time = token["exp"]
+            add_event(f"expiration: {exp_time}")
+            return int(exp_time) - time.time() > 60
+    except Exception:
```
shreyb, line 115: What are your thoughts about having two kinds of `Exception` here, one clause for `scitokens.SciTokensException` and the other for general `Exception`?  That way we could even test both perhaps.

lib/token_mods.py
shreyb: This function could even handle the raising of the right scitokens `Exception` in the case where it's needed.

lib/fake_ifdh.py
```diff
@@ -75,10 +79,14 @@ def getRole(role_override: Optional[str] = None, verbose: int = 0) -> str:
...
-    if os.environ.get("BEARER_TOKEN_FILE", False):
-        with os.popen("decode_token.sh $BEARER_TOKEN_FILE", "r") as f:
-            token_s = f.read()
-            token = json.loads(token_s)
+    if os.environ.get("BEARER_TOKEN_FILE", False) and os.path.exists(
+        os.environ["BEARER_TOKEN_FILE"]
+    ):
+        with open(os.environ["BEARER_TOKEN_FILE"]) as f:
+            token_encoded = f.read(4096).strip()
+            token = dict(
```
retzkek, line 87: Why do you convert the claims to a dict? Why not just call the `get()` method of the scitoken object itself? https://scitokens.readthedocs.io/en/latest/scitokens.html#scitokens.scitokens.SciToken.get

config
shreyb, line 1: Please roll this config update back.

shreyb (changes requested):
I think this is pretty good.  There are some bits that I think could be cleaned up at a future time, but the one thing I'd like to see if we could address now is the broad exception in the case of scitokens parsing/validating failing.  To be clear, I'm okay with the code going in as is, if we want to defer the cleanup to a later version.

lib/fake_ifdh.py
```diff
@@ -75,10 +79,14 @@ def getRole(role_override: Optional[str] = None, verbose: int = 0) -> str:
...
     # if there's a role in the wlcg.groups of the token, pick that
-    if os.environ.get("BEARER_TOKEN_FILE", False):
-        with os.popen("decode_token.sh $BEARER_TOKEN_FILE", "r") as f:
-            token_s = f.read()
-            token = json.loads(token_s)
+    if os.environ.get("BEARER_TOKEN_FILE", False) and os.path.exists(
+        os.environ["BEARER_TOKEN_FILE"]
+    ):
+        with open(os.environ["BEARER_TOKEN_FILE"]) as f:
+            token_encoded = f.read(4096).strip()
```
retzkek, line 86: Why limit to 4096 bytes? That's only 2-3x what a current analysis token is, and we may see that explode (cf. recent request from mu2e). If you must have a limit, at least give us a couple orders of magnitude buffer.

retzkek (commented):
Thanks Marc! A few inline comments.

lib/fake_ifdh.py
```diff
@@ -75,10 +79,14 @@ def getRole(role_override: Optional[str] = None, verbose: int = 0) -> str:
...
+            token = dict(
+                scitokens.SciToken.deserialize(token_encoded, insecure=True).claims()
```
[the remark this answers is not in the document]
marcmengel: I don't want to require that we have all the environment/config for /etc/grid-security/certificates setup so as to
be sure we can validate the certificates, etc.  The services we call need that all setup, but we don't.
Also, trying to replace decode-token.sh, which never validated anything.

lib/fake_ifdh.py
```diff
@@ -75,10 +79,14 @@ def getRole(role_override: Optional[str] = None, verbose: int = 0) -> str:
             return role
 
     # if there's a role in the wlcg.groups of the token, pick that
-    if os.environ.get("BEARER_TOKEN_FILE", False):
-        with os.popen("decode_token.sh $BEARER_TOKEN_FILE", "r") as f:
-            token_s = f.read()
-            token = json.loads(token_s)
+    if os.environ.get("BEARER_TOKEN_FILE", False) and os.path.exists(
```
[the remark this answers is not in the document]
marcmengel: We have our own convention for keeping per-experiment token files (`/tmp/bt_token_${EXPERIMENT}_${ROLE}_${UID}`)  that we share with ifdh, and that's already been resolved town to BEARER_TOKEN_FILE by the time we get here; the SciToken discover method I don't think wants to follow that system.

lib/fake_ifdh.py
marcmengel: We want it to just return False; the scitokens library (and underlying JWT library0 raises different exceptions for expired, bad signature, empty, etc...  Maybe we want a flag passed in that says if we should return false or re-raise?  We could use the former behavior when deciding if we should call htgettoken, and the latter to check the token the second time, so it raises an exception about what was wrong?

retzkek: >  Maybe we want a flag passed in that says if we should return false or re-raise? 

I like this idea.

lib/fake_ifdh.py
marcmengel: The dict bit is to help make mypy happy.   There isn't a typing stub for scitokens (or at least I didn't find one) so
that makes it a Dict[Any,Any] and then we convert the Any  we extract with str() when we fetch it and mypy is happy.
I went about 4 rounds with other permutations trying to get mypy to not  choke on it.

Also, you can't do x['key'] directly on a x = SciToken(), but you can x.claims()['key'] , but you can't do x.claims().get('key')... 
but you can do dict(x.claims()).get("key") AND dict(x.claims)["key"].   At least that's what I remember
from my interactive tests.  So I call claims() and convert it to a dict everywhere, and I can use either
.get(k) or [k] on it.

lib/fake_ifdh.py
marcmengel: I was getting an error earlier on (because I wasn't stripping the newline off the file contents) and I thought I was perhaps getting a short read; so the 4096 was an attempt to be doubly sure it was reading the whole file... I should take that back out...

lib/fake_ifdh.py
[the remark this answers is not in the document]
retzkek: It looks at `BEARER_TOKEN_FILE` second (only after `BEARER_TOKEN`), so should work fine for the jobsub and more general use cases, and would eliminate a few more lines of boilerplate (that @shreyb asked about factoring out above).

lib/fake_ifdh.py
shreyb: @marcmengel  and @retzkek:  So this reraise solution I _don't_ particularly like.  It strikes me as a bit of a code smell and we won't remember six months from now why we did it and why `checkToken` has this random `reraise` bool.

Instead, I do like the original idea of Kevin's that the caller (which is just the two times in `getToken`) knows what to do.  I think it'd keep the intended return states of `checkToken` much clearer.  If we didn't want to change much, we could only change the `getToken` code so it checks for `True` or `False` the first go-around (like it currently does), and the second time, it will do a try-except or something like that.

If we want to get a bit more adventurous, but IMO more clear in the long run, and since the impact is pretty localized, we could even change the signature of `checkToken` so it just returns None, and raises an `Exception` in any error case or a failure case, which `getToken` would know how to handle.  

For example, `checkToken` currently only returns `False` if the tokenfile doesn't exist, or fails the time to expiration check.  Both of these could be raised `Exceptions` (an `OSError` for the first, and maybe a custom `TokenExpirationError` for the second), which `getToken` would know how to handle.  Your current restructuring of the code, Marc, actually allows this to be implemented pretty cleanly.  Then, if `checkToken` raises any other exception, we'd reraise that (or reraise ANY `Exception` in the second case).

I think that'd look something like this (pseudocode):

```python

def checkToken(tokenfile: str) -> None:
    exp_time = None
    try:
        with open(tokenfile) as f:
            do_stuff()
            assert int(exp_time) - time.time() > 60
    except OSError:
        # File not found
        raise
    except AssertionError as e:
        raise TokenExpirationError("Token expires less than 60 seconds from now, at {string_for_time}")
    # Other future exception clauses?
    except Exception as e:
        raise

def getToken(role: str = DEFAULT_ROLE, verbose: int = 0) -> str:
    do_stuff()
    # Change this call since we'd return None
    try:
        checkToken(tokenfile)
    except (OSError, TokenExpirationError) as e:
        # maybe print the exception so in verbose mode we know what's going on?
        if verbose > 0:
            print(e). # Something better than this
        do_the_htgettoken_stuff_including_raising_PermissionError_if_it_fails()
        checkToken(tokenfile)  # This will automatically reraise if something fails this time
        return tokenfile
    except Exception as e:
        raise e
    else:
        return tokenfile
```

This is a long bit of pseudocode, and yes, @retzkek , a little Go-ish.  I'm happy to type it up for real and put it in here if you guys think it's a good idea.  I think it'll make it very clear that `checkToken` either returns nothing if the check works, or some kind of Exception if it fails, and then getToken handles any reraising or not, which satisfies Kevin's initial point (one I agree with).

What are your thoughts?

shreyb (changes requested):
I'd like to propose a restructuring of `checkToken` and how `getToken` calls it to more align with what @retzkek  said originally.  I'm not a huge fan of the `reraise` bool (see the comment for explanation)

lib/fake_ifdh.py
```diff
@@ -131,7 +133,13 @@ def getToken(role: str = DEFAULT_ROLE, verbose: int = 0) -> str:
...
         os.environ["BEARER_TOKEN_FILE"] = tokenfile
 
-    if not checkToken(tokenfile):
+    try:
+        token_ok = checkToken(tokenfile)
+    except:
+        # if anything goes wrong checking, its not okay, get a fresh one
+        token_ok = False
+
+    if not token_ok:
```
shreyb, line 142: I like this a lot better - thanks Marc!  

One last nitpick (but definitely not a blocker for approval IMO):

Given this new logic, what would you think about, on line 154, changing the last `checkToken` check from this:

```python
if checkToken(tokenfile):
    return tokenfile
raise PermissionError("blahblah")
```
to:

```python
if not checkToken(tokenfile):
    raise PermissionError("blahblah")
```

This way:

1. If that second `checkToken` fails, we get the `PermissionError` raised
2. If an `Exception` is raised from `checkToken`, it gets raised anyway, and 
3. If nothing goes wrong, this clause is skipped, and the `return tokenfile` on what is now line 157 gets executed.

I think that small change might make the logic a little clearer given the context starting from line 151, so that we're only returning in that clause if there's an issue.  In context, that'd look like:

```python
    if not token_ok:
        do_the_htgettoken_and_role_stuff()
        res = os.system(cmd)
        if res != 0:
            raise PermissionError(f"Failed attempting '{cmd}'")
        if not checkToken(tokenfile):
            raise PermissionError(f"Failed validating token from '{cmd}'")
    return tokenfile
```

shreyb (changes requested):
I like this a lot better.  I do have one last nitpick though, about the second time we run `checkToken`.

lib/fake_ifdh.py
shreyb: Thanks!

shreyb (approved):
Thanks Marc!  I'm good with this PR.  @retzkek - have all of your concerns been addressed?

lib/fake_ifdh.py
retzkek: > Also, you can't do x['key'] directly on a x = SciToken(), but you can x.claims()['key'] , but you can't do x.claims().get('key')...

You _can_ call `x.get('key')` though.

retzkek:
No, some of my comments are still open. I'd really prefer we leverage the functionality the library provides when possible.

lib/fake_ifdh.py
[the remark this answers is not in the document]
marcmengel: Okay, so I updated the calls to use discover(), except for the case in token_mods where I'm not sure the BEARER_TOKEN_FILE is set to the token file;  It does look a lot cleaner that way.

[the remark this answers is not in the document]
marcmengel: So we're using token discovery now everywhere except one spot in token weakening where we aren't setting BEARER_TOKEN_FILE; I think it looks a lot cleaner now -- @kretzke can you take another look?

lib/fake_ifdh.py
marcmengel: So basically now checkToken always raises exceptions, and in the place we want to keep going if it fails, we do a generic except: that just says we need to fetch a new token.    THen having tried to fetch one, if checkToken raises anything we let that kill us.

lib/fake_ifdh.py
marcmengel: This should all be cleaned up now.

retzkek:
LGTM

marcmengel:
I'll just put in another pull request later to remove decode_token.sh

marcmengel:
Sorry, I ended up just re-adding decode-token.sh, so the merge conflict went away, and submitted
a new pull request to get it back out.

--
Marc W. Mengel
Computer Services Senior Developer

Data Management and Applications
Fermi National Accelerator Laboratory
630 840 8256 office
www.fnal.gov
<https://email.fnal.gov/owa/redir.aspx?C=JkVsWjSyHEuWV2kFJWNPyC9OyMKf7dEIp_yxPk9yfhfadE3mdLfI6T8Ztf8BVaJjzoIe8oJ5H0M.&URL=http%3a%2f%2fwww.fnal.gov>
Connect with us!
Newsletter<https://email.fnal.gov/owa/redir.aspx?C=-29AtpIe40SAvGF4aiNr86KvVjap7dEIuxIdg8Q5bYa9sT7QTFVhxhVnl8zwXkEuBSuAQLRkEfg.&URL=http%3a%2f%2fwww.fnal.gov%2fpub%2ftoday> | Facebook<https://email.fnal.gov/owa/UrlBlockedError.aspx> | Twitter<https://email.fnal.gov/owa/redir.aspx?C=-29AtpIe40SAvGF4aiNr86KvVjap7dEIuxIdg8Q5bYa9sT7QTFVhxhVnl8zwXkEuBSuAQLRkEfg.&URL=https%3a%2f%2ftwitter.com%2fFermilabToday>

shreyb:
No problem - that works.

PR #482, opened 2023-10-04
Cleanups building RPM on alma9...

marcmengel:
Fixing Makefile typos
clearing test teardown failure


shreyb:
There will unfortunately be one more config change after #481  is merged.

shreyb (commented):
LGTM.  Let's hold off until #481 is done, then we can update the config here and merge.

PR #486, opened 2023-10-17
Adding jobsub_cleanup_cred script

marcmengel:
Tool for users who have hosey credentials stuck on the credds to clean them all out so they can resubmit.

### DIFF
--- a/bin/jobsub_cleanup_cred
+++ b/bin/jobsub_cleanup_cred
@@ -1,0 +1,115 @@
+#!/usr/bin/python3 -I
+
+#
+# jobsub_cleanup_cred -- utility to cleanup tokens, etc. on condor_credd's
+#
+# COPYRIGHT 2022 FERMI NATIONAL ACCELERATOR LABORATORY
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+#
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+    utility to cleanup tokens, etc. on condor_credd's
+"""
+# pylint: disable=wrong-import-position,wrong-import-order,import-error
+
+import argparse
+from collections import defaultdict
+import os
+import sys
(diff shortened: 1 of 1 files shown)


bin/jobsub_cleanup_cred
```diff
@@ -0,0 +1,120 @@
...
+sys.path.append(os.path.join(PREFIX, "lib"))
+
+import fake_ifdh
+from creds import get_creds, CredentialSet
+import get_parser
+from condor import get_schedd_names
+
+notes = """
+
+BEARER_TOKEN_FILE=/tmp/bt_token... _condor_COLLECTOR_HOST=dunegpcoll02.fnal.gov _condor_CREDD_HOST=dunegpschedd02.fnal.gov condor_store_cred qery-oauth
```
shreyb, line 45: Should this be `query-oauth` at the end?

bin/jobsub_cleanup_cred
```diff
@@ -0,0 +1,120 @@
...
+
+A credential was stored and is valid.
+
+Credential info:
+dune.top = 1681431891
+dune.use = 1682976510
+dune_ab0f4c39c0.top = 1682090936
+dune_ab0f4c39c0.use = 1682976330
+and:
+BEARER_TOKEN_FILE=/tmp/bt_token... _condor_COLLECTOR_HOST=dunegpcoll02.fnal.gov _condor_CREDD_HOST=dunegpschedd02.fnal.gov condor_store_cred delete-oauth -s credname
```
shreyb, line 59: Please remove the hostnames here.  FNAL security policy, apparently.

bin/jobsub_cleanup_cred
```diff
@@ -0,0 +1,120 @@
...
+        tname, handle = tname.split("_")
+        cmd = f"condor_store_cred delete-oauth -s {tname} -H {handle}"
+    else:
+        cmd = f"condor_store_cred delete-oauth -s {tname}"
+    print(f"Running: '{cmd}' against {schedd}:")
+    os.system(cmd)
+
+
+def main():
+    parser = get_parser.get_jobid_parser(add_condor_epilog=True)
```
shreyb, line 97: Scope increase request:

Could we have a --test or --dryrun argument that would tell the user which credentials on which schedd would get deleted, but not actually do the delete?

shreyb (changes requested):
I generally like this.  Just a couple of comments and one request.

bin/jobsub_cleanup_cred
marcmengel: We can use -n for it and not add another flag... yes.

shreyb (approved):
LGTM

PR #487, opened 2023-10-18
cleaning out decode_token.sh now that it is no longer used

marcmengel:
This was initially in #480, but caused a merge conflict...


shreyb (approved):
LGTM

Issue #488, opened 2023-10-27
Log cvmfs info at end of job


marcmengel:
From @DrDaveD:

If it isn't too difficult to do, it would be awfully useful if a job wrapper or the pilot or something could collect cvmfs logs for all the repositories that are known about for a failed job.  That can be done through a command like this:
```
attr -g logbuffer /cvmfs/<reponame>|tail -50
```

marcmengel:
We can log this info for every job with "ifdh log", so it can be looked  at in kibana, etc.

PR #489, opened 2023-10-27
Log cvmfs info

marcmengel:
Addition to standard wrapper to log recent cvmfs info, per #488 . 
The cvmfs information can be found in kibana in the fife_dh* logs.
[example](https://landscape.fnal.gov/kibana/app/kibana#/discover?_g=(refreshInterval:(pause:!t,value:0),time:(from:'2023-10-27T22:00:00.000Z',mode:absolute,to:'2023-10-27T23:00:00.000Z'))&_a=(columns:!(_source),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'fife-dh-*',key:query,negate:!f,type:custom,value:'%7B%22term%22:%7B%22ifdh_log_msg%22:%22cvmfs%22%7D%7D'),query:(term:(ifdh_log_msg:cvmfs))),('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'fife-dh-*',key:query,negate:!f,type:custom,value:'%7B%22term%22:%7B%22ifdh_log_msg%22:%22catalog%22%7D%7D'),query:(term:(ifdh_log_msg:catalog)))),index:'fife-dh-*',interval:auto,query:(language:lucene,query:''),sort:!('@timestamp',desc)))

Also throwing in an unrelated script I was adding to the testing area for setting up metacat/rucio data files...


templates/simple/simple.sh
shreyb, line 1: I thought we wanted to log cvmfs lines from failed jobs, not all jobs (referring to SNOW RITM1905386).  

Also, do we have kibana logs from any tests of this new feature?

tests/job_scripts/rebuild_gen_cfg_metacat
shreyb, line 1: Is this test adding to one of our test scripts?  Where does this get used?  Also, does this add a dependency, or does setting up `hypotcode` do that for us?

shreyb (changes requested):

templates/simple/simple.sh
marcmengel: Well, the feature request was for failed jobs; I thought perhaps doing it for all jobs might be good. 
Because even if the job doesn't FAIL, people might want to know what cvmfs revision the job rand with....

The "example" link in the description should show the test logs in kibana.

tests/job_scripts/rebuild_gen_cfg_metacat
marcmengel: I'm planning to add a submission test that does a data-dispatcher get-next-file job, as well as the SAM ones we have now.   This script is  to setup the dataset of files for such a job to read through, much like the one we have there for SAM already. Having it be run is a pre-requisite for adding such a test case; but we don't actually run the SAM one either as part of the tests (although maybe we should)...  So I had it sitting there when I made the branch and thought this was as good a time as any to add it...

shreyb: Fair enough.

templates/simple/simple.sh
shreyb: My worry is about this becoming a storage issue on the monitoring side though, if we're sending 50 records from each job.  Otherwise, I agree with you. 

What are your thoughts @retzkek ?

retzkek:
Before this gets merged I'd like to consider the implications this could have on our logging infrastructure.

templates/simple/simple.sh
marcmengel: So far, I've never seen that log request give more than 5 entries.  We could easily set it to limit to 10 or so, I think; we could ping @DrDaveD to see what led him to request 50 initially.

DrDaveD: I didn't consider storage issues, but sometimes if there are networking issues it can cycle through a lot of proxies and servers and I figured that up to 50 lines might be interesting.  How about collecting just 5 lines for successful jobs and 50 for failed?  Or if you just want revision number, that can be read instead from the `revision` attribute.

retzkek: Looking at the past month, we see about 500K jobs per day, and 20M IFDH messages. So we'd be looking at up to doubling the IFDH log rate and using up to a couple more TiB for 90 days of history. This isn't a deal breaker, it should be manageable, especially if we reduced the retention (90 days does seem a bit much tbh).

@DrDaveD how do you envision this data being used? Do you see value in having logs from multiple machines/sites/jobs together in one place (like Marc's example link)?  What if we only logged the revision with IFDH, but then put the logs into a file in the job sandbox in dCache?

DrDaveD: > @DrDaveD how do you envision this data being used? Do you see value in having logs from multiple machines/sites/jobs together in one place (like Marc's example link)?

Yes I do see value in that.  If we have a lot of jobs failing due to cvmfs errors we will want to look at those logs to find out the cause of those failures.

> What if we only logged the revision with IFDH, but then put the logs into a file in the job sandbox in dCache?

I don't understand the implications.  If it's easy enough to look into the cvmfs logs of jobs that fail due to cvmfs errors I'm not sure what difference it makes.

shreyb:
The only issue I see with sending the cvmfs log to the sandbox is that some
folks (like I believe @DrDaveD) won’t be able to see them without us
downloading them for him.  Otherwise, I think it would work well.

templates/simple/simple.sh
retzkek: OK I'm fine with sending the logs through IFDH.

marcmengel: I think so?  So basically full log( max 50 lines ) on failure, last revision num on success?   I think that's pretty easy.

marcmengel: That's what I've pushed now.   We log the last cvmfs version line on successful jobs, and all available (up to 50) on failed jobs (exiting non-zero).

shreyb: Cool - do you have a test job (maybe one success, one failed) we could take a look at in Kibana to see what this looks like?

shreyb: It looks good to me otherwise - but I just want to make sure what we expect is what we actually see.

marcmengel: See https://fifemon.fnal.gov/kibana/goto/141c1ddba3bd4eb7f4a06d6f6ba53f1c

marcmengel: 5806269.0@jobsub04.fnal.gov was a /bin/true job, 5704583.0@jobsub05.fnal.gov was a /bin/false job ("falied")

shreyb: Thanks!  Last nitpicky thing here.  Since the record already has the `JobsubJobId`, do we need to repeat that in the log message?  

Instead, what do you think of explicitly stating that the job is completed at the time of the logging?  Something like `"Completed job cvmfs: $line"` (or anything else you think is helpful and brief).

DrDaveD:
As an optimization, anything that's as old or older than the second-oldest `switched to catalog revision` message could be dropped because it will not be relevant to any cvmfs failure.

shreyb:
> As an optimization, anything that's as old or older than the second-oldest `switched to catalog revision` message could be dropped because it will not be relevant to any cvmfs failure.

This is a good idea.  I leave it to you, @marcmengel, to decide whether or not to have it go into this PR or split it into another PR/issue/release.

marcmengel:
Urf.  That took slightly more tries to get right than it should have, but I think this is a good version. 
So we give the last revision update log line if we're successful, but if not, we check if there is more than one
revision update log line, and take everything after the second-to last one if there is; if there are not more than one revision update log line, we take the whole log.

shreyb:
My good job test looks good (5957618.0@jobsub05.fnal.gov), but my `/bin/false` test (5977824.0@jobsub04.fnal.gov) doesn't show any cvmfs revision lines.  It's possible the job ran for so short that it didn't see any of those.  Let me make a test script that sleeps 5 minutes and then fails, and see if we get better info there.

marcmengel:
There are some nodes for which that loginfo command returns nothing, but I usually haven't had to run more than 2 submissions to get one with something in the logs.

shreyb:
Good call on that.  My latest job had that issue.  I'll force `--onsite-only` so that we get an FNAL node for that test.

shreyb:
Hmm - I'm not able to get the revisions for the failed jobs.  Do you have an example test that you ran?

marcmengel:
https://landscape.fnal.gov/kibana/goto/271acb0cb3e82133ad201d0dd4c2163d

marcmengel:
Oh, the other hint is to run -G dune when doing these tests, the fermilab cvmfs is really boring.

marcmengel:
Also, the stderr from that job is:
```
Mon Nov 6 20:01:10 UTC 2023 5952763.0@jobsub04.fnal.gov BEGIN EXECUTION false
cvmfs info:
Attribute "logbuffer" had a 504 byte value for /cvmfs/dune.opensciencegrid.org/:
[6 Nov 2023 11:52:23 UTC] switching proxy from (none) to http://192.12.238.153:3
128 (set proxies)
[6 Nov 2023 11:52:23 UTC] geographic order of servers retrieved from cvmfs-s1bnl
.opensciencegrid.org
[6 Nov 2023 11:52:23 UTC] switching proxy from http://192.12.238.153:3128 to htt
p://192.12.238.152:3128 (geosort)
[6 Nov 2023 11:52:23 UTC] switching proxy from (none) to http://192.12.238.154:3
128 (cloned)
[6 Nov 2023 11:52:23 UTC] switching proxy from http://192.12.238.154:3128 to DIR
ECT (set proxies)
Mon Nov 6 20:01:11 UTC 2023 false COMPLETED with exit status 1
```

shreyb:
Alright - then let's go ahead and merge this.

PR #491, opened 2023-11-01
Fake test tokens expire in 2222-02-22 22:22:22 GMT

shreyb:
The tokens I added yesterday in the tests expired, so unit tests started failing.  I added new fake tokens generated from https://demo.scitokens.org that expire 2222-02-22 22:22:22 GMT.  Presumably we'll have something newer than JWT by then.


marcmengel (approved):
LGTM

Issue #493, opened 2023-11-08
Add some debug lines


shreyb:
@njpfnal requests that in our `--debug` mode we should add the following info:

Timestamp
[Submit] hostname

PR #496, opened 2023-11-08
log host and time 

marcmengel:
log host and time to stderr if verbose/debug and to Jaeger trace, per #493

### DIFF
--- a/bin/jobsub_cmd
+++ b/bin/jobsub_cmd
@@ -39,7 +39,7 @@ import creds
 import get_parser
 import condor
 from version import print_version, print_support_email
-from tracing import as_span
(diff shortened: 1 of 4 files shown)


shreyb (changes requested):
`args.verbose` is parsed from the args as an `int` (see `get_parser.py`).  So we either should convert it to a bool somehow before calling `log_host_time`, or have `log_host_time` accept `verbose` as an int.

Given python's typing system, it most likely would work as is, but in the interest of keeping the code cleaner, I'd prefer we stay consistent with the typing of vars we're passing around.

Otherwise, it looks good.

shreyb (approved):
LGTM

PR #499, opened 2023-11-17
move scitokens cache to /var/run

marcmengel:
Since they didnt' bother to make a specific environment variable for it or whatever... sigh.

### DIFF
--- a/lib/fake_ifdh.py
+++ b/lib/fake_ifdh.py
@@ -41,6 +41,20 @@ VAULT_OPTS = htcondor.param.get("SEC_CREDENTIAL_GETTOKEN_OPTS", "")
 DEFAULT_ROLE = "Analysis"
 
 
+def init_scitokens() -> None:
+    cfgfile = f'{os.environ.get("HOME", "/tmp")}/.jobsub_scitokens.cfg'
+    uid = os.getuid()
+    if not os.access(cfgfile, os.R_OK):
+        with open(cfgfile, "w") as cff:
+            cff.write(
+                f"[scitokens]\ncache_location: /var/run/user/{uid}/jobsub_scitokens.cache"
+            )
+    scitokens.set_config(cfgfile)
+
+
+init_scitokens()
+
+
 def getTmp() -> str:
     """return temp directory path"""
     return os.environ.get("TMPDIR", "/tmp")


lib/fake_ifdh.py
```diff
@@ -41,6 +41,20 @@
 DEFAULT_ROLE = "Analysis"
 
 
+def init_scitokens() -> None:
+    cfgfile = f'{os.environ.get("HOME", "/tmp")}/.jobsub_scitokens.cfg'
+    uid = os.getuid()
+    if not os.access(cfgfile, os.R_OK):
+        with open(cfgfile, "w") as cff:
+            cff.write(
+                f"[scitokens]\ncache_location: /var/run/user/{uid}/jobsub_scitokens.cache"
```
shreyb, line 50: Two things:

1.  The cache_location is supposed to be a directory where the DB lives, not the file itself (see the table here:  https://github.com/scitokens/scitokens#configuration)

2.  Just to make this more readable for future us, what are your thoughts about taking advantage of python's multiline strings and making this write statement:

```python
cff.write((
    "[scitokens]\n"
    f"cache_location: /var/run/user/{uid}\n"
))
```
(I added a newline at the end of the `cache_location` designation just o make the file look cleaner if someone `cat`s it in the future).

retzkek:
Is it guaranteed that  `/var/run/user/{uid}/` will exist? Should at least check, no?
Looks like that's a tmpfs filesystem, I guess that's OK? Cache is small, and if it has to be recreated occasionally so be it.
Also I notice on an AL9 system that it creates `/run/user/{uid}` instead. 
Ain't it fun?

lib/fake_ifdh.py
```diff
@@ -41,6 +41,20 @@
 DEFAULT_ROLE = "Analysis"
 
 
+def init_scitokens() -> None:
```
shreyb, line 44: Would you mind putting a docstring in there that explains 1) that we're doing this to NFS-proof the key-caching, and 2) That we're dynamically generating a config-file that the scitokens library will use?  I just want to make sure that six-months-from-now-us will know what this is even for.

shreyb (changes requested):

shreyb:
To @DrDaveD's point in slack, and @retzkek's here, since `/var/run/user` may not even exist, depending on OS, we're already dumping small files (tokens) into /tmp, we could even dump this in there too.  Maybe make a directory in there for the user, something like `/tmp/scitokens_$(id -u)`?  So you'd make that dir if it doesn't exist, and then set `cache_location` there?

Before merging this PR, we should make sure that we test this both with regular and shared-account users.

retzkek:
We _could_ even theoretically let users share a single cache file in /tmp, but that's probably just inviting more trouble.

DrDaveD:
`$XDG_RUNTIME_DIR` is set to `/run/user/{uid}` even on el7; `/var/run` is a symlink pointing to `/run`.  The directory and the environment variable are created only by systemd for interactive logins.  That directory cannot be created by an unprivileged program.  The backup will need to be something either under `/tmp` (as is the case for the WLCG bearer token discovery protocol) or `/dev/shm/`.

DrDaveD:
In any case it would be better for a fix to be in the scitokens library instead of in jobsub, since that's where the problem is.

marcmengel:
So, just to avoid cases where the runtime dir isn't set, etc.  I've simplified it to put the scitokens config and the token cache all in the same /tmp/js_scitoken_{UID} directory (with $TMPDIR overriding /tmp)   Then there is only one directory to make and manage.   I also have it touch the config file whenever we run, so /tmp cleaners don't delete it out from under us.

Added a doc strng as well, and passes unit tests.

marcmengel:
So it looks like this:
```
[mengel@fifeutilgpvm01 lib]$ ls -laR /tmp/js_scitok_1733
/tmp/js_scitok_1733:
total 128
drwxrwxr-x   3 mengel mengel   4096 Nov 20 11:25 .
drwxrwxrwt. 16 root   root   114688 Nov 20 11:29 ..
drwxrwxr-x   2 mengel mengel   4096 Nov 20 11:25 scitokens
-rw-rw-r--   1 mengel mengel     48 Nov 20 11:28 .scitokens.cfg

/tmp/js_scitok_1733/scitokens:
total 12
drwxrwxr-x 2 mengel mengel 4096 Nov 20 11:25 .
drwxrwxr-x 3 mengel mengel 4096 Nov 20 11:25 ..
-rw-r--r-- 1 mengel mengel 3072 Nov 20 11:25 scitokens_keycache.sqllite
[mengel@fifeutilgpvm01 lib]$ cat /tmp/js_scitok_1733/.scitokens.cfg
[scitokens]
cache_location: /tmp/js_scitok_1733
```

marcmengel:
But I agree with @DrDaveD, it should be fixed upstream.   Putting the token cache in $HOME is a problem with NFS mounted $HOME areas, and that should be adressed up there.   In the mean time, however, this is a workaround that
should let us keep running.

marcmengel:
And I see that @retzkek beat me to it already https://github.com/scitokens/scitokens/issues/188.

marcmengel:
So in case this moving to /tmp thing *doesn't* fix the bug, remove zero length cache files, if found.

lib/fake_ifdh.py
shreyb: I like the new edit.  Did you think we should do something like point (2) here, or do you think it's unnecessary?

shreyb (commented):
I have one open conversation point - definitely not a blocker IMO - just a stylistic question.

DrDaveD:
Alternatively it could set only `$XDG_CACHE_HOME` if it isn't set, as suggested in https://github.com/scitokens/scitokens/issues/188#issuecomment-1820558467.

shreyb:
> Alternatively it could set only `$XDG_CACHE_HOME` if it isn't set, as suggested in [scitokens/scitokens#188 (comment)](https://github.com/scitokens/scitokens/issues/188#issuecomment-1820558467).

I like this idea, but for now, since nobody has any functional issues with Marc's PR, let's merge it and get this released so users won't have to do the workaround themselves.

I'll open a new issue with this suggestion so we can try it out, since it _is_ much cleaner than our workaround.  I suppose we can set `$XDG_CACHE_HOME` for only the jobsub operation (and then put it back), but I want to make sure we're not breaking anything else.

PR #502, opened 2023-11-27
Cleanup tarfiles

marcmengel:
Basically, some of our cleanup calls weren't being made if we got exceptions at various points.
A couple of try:...finally: blocks should be the remedy.

Reviewers: look at the diffs ignoring whitespace, as blocks of text were indented for the try:finally: bits.


shreyb (approved):
Looks good to me.

PR #504, opened 2023-12-05
Added capability to Makefile to create release candidates

shreyb:
As it says above.  The idea is just to set the version, RC, and run `make` like normal.

To create a final release, as it says in the `Makefile`, set the RC to `0`.


shreyb:
Forgot to add - I tested this creating both an "RC" and a "Final" release, on both SL7 and AL9.

marcmengel (approved):

PR #508, opened 2023-12-19
Fixup condor submit dag

marcmengel:
The condor_submit_dag wrapper never *really* worked; it now actually adds
all the jobsub variables to the dagman submission, etc; for what that's worth.
Now works generally, and also on AL9.

Refactored "render_files" out of submit_support to prevent cyclic dependency.

A whole bunch of files added for testing, mainly files from a --dataset submission
with a few tweaks (to pass a fresh $SAM_PROJECT...)


shreyb:
Would you be able to fix some of the pre-commit checks?  Especially the mypy errors?

marcmengel:
Ack!  Sorry, I need to get pre-commit setup on my AL9 node...

templates/dag/dagman_wrapper.sh
```diff
@@ -5,6 +5,5 @@
 
 
 # touch our transfer files so condor will copy them back
-(sleep 1; touch {%for f in transfer_files%}{{f}} {%endfor%}) &
```
shreyb, line 8: Sorry if it's my ignorance here, but what does the dash at the beginning of the command do?

marcmengel: That's just the minus in the diff. for the line being deleted.

We don't need this touch stuff anymore, as we are transferring all those files to dcache at submission time;  the touch was so the files would get transferred back by condor_transfer_data...

shreyb: OMG.  I feel so dumb.  Time to look away from the screen for a bit.

lib/render_files.py
```diff
@@ -0,0 +1,77 @@
...
+import glob
+import os
+import os.path
+import sys
+from typing import Union, List, Dict, Any
+from tracing import as_span
+
+import jinja2 as jinja  # type: ignore
+
+""" Code factored out of jobsub_submit """
```
shreyb, line 11: yay!

lib/utils.py
```diff
@@ -100,6 +105,61 @@ def backslash_escape_layer(argv: List[str]) -> None:
         argv[i] = re.sub(r"\\(.)", "\\1", argv[i])
 
 
+def set_some_extras(
```
shreyb, line 108: Small comment.  Is this ONLY needed for DAGs?  If so, could the function name be changed to `set_dag_extras` or something along those lines?

shreyb: On second thought - this seems to be required for DAGs, but for regular jobs, other settings are needed.  If so, ignore my last comment.

marcmengel: Right, I'm just partitioning it to a subset of things we need in condor_submit_dag, then there's a longer list for jobsub_submit, where we fix disk units, etc.   Maybe "some" is a poor choice... sigh.

tests/data/condor_submit_dag/dagbegin.cmd
```diff
@@ -0,0 +1,66 @@
...
++RUN_ON_HEADNODE= True
+rank               = Mips / 2 + Memory
+job_lease_duration = 3600
+transfer_error     = True
+transfer_executable= True
+when_to_transfer_output = ON_EXIT_OR_EVICT
+transfer_output_files = .empty_file
+request_memory = 100mb
+
++JobsubClientDN="/DC=org/DC=cilogon/C=US/O=Fermi National Accelerator Laboratory/OU=People/CN=Marc Mengel/CN=UID:$ENV(USER)/CN=3752395701"
```
shreyb, line 23: Is there a way of not having this DN here (or anonymizing it), or will that cause the job to fail?  Given the `CN` bit at the end, I'm assuming it should be OK to make it be a random string, but please let me know if that's wrong.

tests/data/condor_submit_dag/dagbegin.cmd
```diff
@@ -0,0 +1,66 @@
...
+
+# Credentials
+
+
+use_oauth_services = fermilab
+fermilab_oauth_permissions_b355f5a23c = " compute.read compute.create compute.cancel compute.modify storage.read:/fermilab/users/$ENV(USER) storage.create:/fermilab/users/$ENV(USER) storage.create:/fermigrid/jobsub/jobs "
+
+
+
++x509userproxy = "x509up_fermilab_Analysis_1733"
```
shreyb, line 62: Here we should make sure the UID is that of the testing user.

tests/data/condor_submit_dag/dagend.cmd
```diff
@@ -0,0 +1,67 @@
...
+transfer_executable     = True
+rank               = Mips / 2 + Memory
+job_lease_duration = 3600
+transfer_error     = True
+transfer_executable= True
+transfer_output_files=.empty_file
+when_to_transfer_output = ON_EXIT_OR_EVICT
+request_memory = 100mb
+
++JobsubClientDN="/DC=org/DC=cilogon/C=US/O=Fermi National Accelerator Laboratory/OU=People/CN=Marc Mengel/CN=UID:$ENV(USER)/CN=3752395701"
```
shreyb, line 24: Same comment as above.

tests/data/condor_submit_dag/dagend.cmd
```diff
@@ -0,0 +1,67 @@
...
+
+# Credentials
+
+
+use_oauth_services = fermilab
+fermilab_oauth_permissions_b355f5a23c  = " compute.read compute.create compute.cancel compute.modify storage.read:/fermilab/users/$ENV(USER) storage.create:/fermilab/users/$ENV(USER) storage.create:/fermigrid/jobsub/jobs "
+
+
+
++x509userproxy = "x509up_fermilab_Analysis_1733"
```
shreyb, line 63: Same comment as above.

tests/data/condor_submit_dag/simple.cmd
```diff
@@ -0,0 +1,96 @@
+
+# generated by jobsub_lite
+# debug
+universe           = vanilla
+executable         = simple.sh
+arguments          = --find_setups --prescript-unquote echo%20JOBSUBJOBSECTION%20is%20%24%7BJOBSUBJOBSECTION%7D --setup-unquote hypotcode%20v1_1 --setup-unquote ifdhc%20v2_6_10%2C%20ifdhc_config%20v2_6_15 --self_destruct_timer 700 --debug --getconfig --limit 1 --appvers v1_1 --metadata_extractor hypot_metadata_extractor --addoutput gen.troot --rename unique --dest /pnfs/fermilab/users/mengel/dropbox --add_location --declare_metadata --addoutput1 hist_gen.troot --rename1 unique --dest1 /pnfs/fermilab/users/mengel/dropbox --add_location1 --declare_metadata1 --exe hypot.exe -- -o gen.troot -c hist_gen.troot
```
shreyb, line 6: Will these tests not work if they don't write out to your `/pnfs/fermilab` area?

tests/data/condor_submit_dag/simple.cmd
```diff
@@ -0,0 +1,96 @@
...
+transfer_executable= True
+transfer_input_files = fife_wrap
+# if transfer_output_files is not explicitly set, condor will transfer ALL the files the job touches (unless in grid universe)
+transfer_output_files = .empty_file
+when_to_transfer_output = ON_EXIT_OR_EVICT
+
+request_memory = 500.0
+request_disk = 102400.0KB
+
++JobsubClientDN="/DC=org/DC=cilogon/C=US/O=Fermi National Accelerator Laboratory/OU=People/CN=Marc Mengel/CN=UID:$ENV(USER)/CN=3752395701"
```
shreyb, line 31: Same comment as in last test cmd file.

tests/data/condor_submit_dag/simple.cmd
```diff
@@ -0,0 +1,96 @@
...
+
+use_oauth_services = fermilab
+
+fermilab_oauth_permissions_b355f5a23c = " compute.read compute.create compute.cancel compute.modify storage.read:/fermilab/users/$ENV(USER) storage.create:/fermilab/users/$ENV(USER) storage.create:/fermigrid/jobsub/jobs "
+
+
+
+
+
++x509userproxy = "x509up_fermilab_Analysis_1733"
```
shreyb, line 91: Same comment as in last cmd test file.

shreyb (changes requested):
There are still some places where either your username, name, or UID show up.  Could you please take a look and see if these can be anonymized or generalized?

scripts/fife_launch_test
```diff
@@ -22,7 +22,7 @@ jobsub_submit \
     --find_setups \
     --export-unquote 'TIMESTAMP%3D%24%28date%20%2B%25Y%25m%25dT%25H%25M%25S%29' \
     --addoutput '*.txt' \
-    --dest '/pnfs/nova/scratch/users/mengel' \
+    --dest '/pnfs/nova/scratch/users/'$USER \
```
shreyb, line 25: Is this a typo?  With `'$USER`?  Should it be `$USER'`?

tests/job_scripts/rebuild_gen_cfg
```diff
@@ -35,9 +35,9 @@ do
     {
      "file_name": "c47fe3af-8fdb-4a5a-a110-3f3d52f3cfea-a.fcl",
      "create_date": "2016-10-11T15:49:15+00:00",
-     "user": "mengel",
+     "user": "$USER",
      "update_date": "2018-09-11T15:30:42+00:00",
-     "update_user": "mengel",
+     "update_user": "$USERl",
```
shreyb, line 40: Typo?

shreyb (changes requested):
Looking better.  There are still a couple of "Marc Mengel"s in there, and I think a couple of typos.  Please take a look at the open conversations/comments and see what you think.

Thanks!

scripts/fife_launch_test
marcmengel: No I think that's right; the single quotes would prevent $USER from being expanded, so we just splice it on the end.

tests/job_scripts/rebuild_gen_cfg
marcmengel: Yep.  dropped the l.

tests/data/condor_submit_dag/dagbegin.cmd
marcmengel: Sigh.  Yes.   MOdifying the test to read the whole DN out of the proxy and put it in $DN in the environment....

tests/data/condor_submit_dag/dagend.cmd
marcmengel: Same, setting DN in the environment to pass it in...

scripts/fife_launch_test
shreyb: Ah - I see your point.  Looks good then.

tests/test_submit_wait_int.py
```diff
@@ -206,6 +206,45 @@ def run_launch(cmd, expected_out=1, get_dir=False):
...
+    def __exit__(self, a, b, c):
+        os.chdir(self.returnto)
+
+
+def condor_dag_launch(dagfile, extra=""):
+    """launch a dag from our dag test area"""
+
+    # need some environment variables to fill in the submit files...
+    proxy = fake_ifdh.getProxy("Analysis")
+    with os.popen(f"openssl x509 -subject -noout -in {proxy}") as subjin:
```
shreyb, line 230: Nice.

tests/data/condor_submit_dag/simple.cmd
shreyb: This appears to be fixed now.

shreyb (commented):
This looks good now.  Before we merge it, I'd like to check out the code tomorrow morning and run the tests to make sure there's no hidden gotchas that depend on your environment.

shreyb (approved):
All tests pass.  Approving.

PR #509, opened 2023-12-19
BEARER_TOKEN_FILE checks for experiment and role

shreyb:
If BEARER_TOKEN_FILE is set to an invalid token or a valid token for the wrong experiment, fail submission with error.

The behavior of this PR enforces the following checks when BEARER_TOKEN_FILE is set:

1.  If the token is expired, we generate a new one at BEARER_TOKEN_FILE, like before
2. If the token has the wrong group or role (as determined by the `-G` and `--role` flags), then we raise a ValueError and let the user know that their token has the wrong role.
3. If the token is malformed somehow and doesn't have a `wlcg.groups` claim, or one without a list, we raise a TypeError (since `None` and `!List` are the wrong types to be stored in that claim)
4. If the token is entirely malformed, we tell the user that

This PR also contains numerous new unit tests to check the above points.


shreyb:
Extra note:  I tested this with my fermilab/Analysis user as well as the dunepro managed token (group dune, role production)

PR #511, opened 2024-01-03
Don't assume we have GROUP set in the environment

### LINKED ISSUE #507: Don't assume that we have the GROUP set in the environment
https://github.com/fermitools/jobsub_lite/blob/da61f1a11190e06bc55c31cbeb41ee7755addda7/lib/fake_ifdh.py#L113

When running unit tests for the 1.5.1 release, I realized that we assume in `getRole` that the `GROUP` will be set.  Now, it should be, due to how the parser works, but it seems bad practice to just make that assumption in this function.  Perhaps we should have a `group_override` arg to the function, and default to that, else check the environment?  Something like that.

The corresponding unit test, 

https://github.com/fermitools/jobsub_lite/blob/da61f1a11190e06bc55c31cbeb41ee7755addda7/tests/test_fake_ifdh_unit.py#L48

Should also set the group for the duration of the test, so it won't error out.  We should also consider defining behavior both in the function and in the test for if a group is NOT defined (probably error out?)

shreyb:
Closes #507 

The main changes in this PR are as follows:

1.  If GROUP is not set in the environment, `getRole` calls `getExp` to get the group.  Previously, we would run into a `KeyError` in this case. 
2. Refactored `getRole` by extracting two functions that explain what we're checking to get the role:  `getRole_from_default_role_file` and `getRole_from_valid_token`.  This cleans up the `getRole` code and makes the logic much clearer.
3. Fixed bug in `getExp` where we didn't strip the newline character from the fallthrough `id -gn` case, and added a test for that case.
4. Lots of unit test refactoring in `test_fake_ifdh.py` to group the tests into classes, since we (rightly so) have multiple tests per `fake_ifdh.py` function.  I also parametrized tests where it made sense.

### DIFF
--- a/lib/fake_ifdh.py
+++ b/lib/fake_ifdh.py
@@ -128,15 +128,13 @@ def getExp() -> str:
 @as_span("getRole")
 def getRole(role_override: Optional[str] = None, verbose: int = 0) -> str:
     """get current role"""
-
     if role_override:
         return role_override
 
     # if we have a default role pushed with a vault token, or $HOME/.jobsub_default... use that
     uid = os.getuid()
-    # sometimes we're called before parsing args, so GROUP may not be
-    # set yet...
-    group = os.environ.get("GROUP", "unknown")
+    environ_group = os.environ.get("GROUP")
+    group = environ_group if environ_group else getExp()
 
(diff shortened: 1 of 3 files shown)


shreyb:
Fixed merge conflicts.

marcmengel (approved):
Looks good.

PR #517, opened 2024-01-17
fixes #516

### LINKED ISSUE #516: jobsub_q bug
This is from INC000001166443 opened by @hgreenlee 
```
Outage start date: 2024-01-17 10:43:57 (US/Central)
Outage/Incident details:
jobsub_q intmittently producing the following diagnostic output.

$ jobsub_q > /dev/null
$ jobsub_q > /dev/null
  File "/opt/jobsub_lite/bin/jobsub_totals", line 28, in <module>
    totals[words[5]] += 1
IndexError: list index out of range
```

Doing some investigation I got the following reproducer of the issue:
```
echo "7597670.0@jobsub05.fnal.gov uboonepro 01/17 12:38 0+00:00:00 R 0 0.0 merge-ly_rayleigh-prod_numi_nu_overlay_run4_fhc_wirecell_detvar-v08_00_00_64.sh --group uboone
All queues are empty" | /opt/jobsub_lite/bin/jobsub_totals
```
the produced output is:
```
7597670.0@jobsub05.fnal.gov uboonepro 01/17 12:38 0+00:00:00 R 0 0.0 merge-ly_rayleigh-prod_numi_nu_overlay_run4_fhc_wirecell_detvar-v08_00_00_64.sh --group uboone
All queues are empty
File "/opt/jobsub_lite/bin/jobsub_totals", line 28, in <module>
totals[words[5]] += 1
IndexError: list index out of range
```

Here the relevant line causing the issue seems to be: `All queues are empty`, that has 4 fields, after a (long) list of lines with more than 4 fields.

marcmengel:
Fix for #516

### DIFF
--- a/bin/jobsub_totals
+++ b/bin/jobsub_totals
@@ -24,7 +24,7 @@ totals["T"] = 0
 for line in sys.stdin.readlines():
     print(line, end="")
     words = re.split(r"\s+", line)
-    if len(words) > 4:
+    if len(words) > 5:
         totals[words[5]] += 1
     totals["T"] += 1
 
--- a/bin/jobsub_totals
+++ b/bin/jobsub_totals
@@ -23,10 +23,10 @@ totals["T"] = 0
 
 for line in sys.stdin.readlines():
     print(line, end="")
-    words = re.split(r"\s+", line)
-    if len(words) > 5:
-        totals[words[5]] += 1
-    totals["T"] += 1
+    m = re.match(r"\d+\.\d+@\S+(\s+\S+){4}\s+([CXIRHS])", line)
+    if m:
+        totals[m.group(2)] += 1
+        totals["T"] += 1
 
 for i in type_map:
     print(totals[i], type_map[i], end="")


bin/jobsub_totals
```diff
@@ -24,7 +24,7 @@ totals["T"] = 0
 for line in sys.stdin.readlines():
     print(line, end="")
     words = re.split(r"\s+", line)
-    if len(words) > 4:
+    if len(words) > 5:
```
shreyb, line 27: So this certainly works, but do you think it'd be more robust to flag the lines based on a regex rather than a line that's longer than `n` words?  It could even be as simple as looking for lines that start with a `\d+\.\d+@.*` to make sure we're starting with a jobid (I totally threw together that regex - if you do use this suggestion, please check that it actually works).

I only worry that if condor changes its output format in the future, this might come back to bite us.

shreyb (commented):
So this certainly works, but do you think it'd be more robust to flag the lines based on a regex rather than a line that's longer than n words? It could even be as simple as looking for lines that start with a `\d+\.\d+@.*` to make sure we're starting with a jobid (I totally threw together that regex - if you do use this suggestion, please check that it actually works).

I only worry that if condor changes its output format in the future, this might come back to bite us.

bin/jobsub_totals
shreyb: Moved comment to general review comment section

marcmengel:
Well, we're doing a condor_q -format in this case, so we *should* only be getting lines in the format we want, or rarely, the "All queues are empty"  line.   The only way we get a problem is if they made the 5th word of a future "All queues are empty" line one of the status letters (TCXIRH), which would then bump the counter for the total of that type... I guess the thing that might fix, upon re-reading it,  is we might be counting an "All queues are empty" line as a job in our total number of jobs.  (i.e. if we had nothing running on any of the 4 servers, we might report 4 jobs total...)     Okay, you talked me into it .

shreyb (approved):
Ooh yes - this is even better than what I suggested.  LGTM.

PR #520, opened 2024-01-24
Improvements to TarfilePublisherHandler

### LINKED ISSUE #468: Improvements to TarfilePublisherHandler
@DrDaveD and I were troubleshooting an imbalance on publishes between the RCDS servers, and he had a couple of suggestions with how we handle retries in the `TarfilePublisherHandler`

https://github.com/fermitools/jobsub_lite/blob/1cf07430b89cfb3e008d32f3b5d5e8d9fff0b8e5/lib/tarfiles.py#L446

# Idea 1

We basically do three operations maximum:  Exists query, Publish request (if not Exists), and then Exists queries (in retry loop until we exhaust retries).  Currently, we switch between servers for each one, and there's not really a reason to do so.  This also obscures any potential issue with a particular server, since we keep switching.  So between the first Exists, and Publish, it'd be nice to just stick with one server unless there's an actual exception.  Between the Publish and the next Exists (and any following retries), we can switch servers.

# Idea 2

For any retries, since we have multiple servers, we currently wait for 30 seconds between retries, even on first failure.  It'd be better if for the first failure, we *immediately* switch servers and retry, and then if there's a second failure, then wait the 30 seconds thereafter between retries.  This should be as easy as doing something like this in the `pubapi_operation` retry loop:

```python
_retry_interval_sec = 0
retry_count = itertools.count()
while True:
  try:
      request()
  except:
      switch_servers()
      rest_of_stuff()
      time.sleep(_retry_interval_sec)
      _retry_interval_sec = RETRY_INTERVAL_SEC
```

We do something similar to this in the tarball discovery code in the job wrapper scripts, where we check all four repositories for a tarball hash, then if we don't find the tarball, we wait 30 seconds and try again.

shreyb:
Closes #468 

This PR has a few changes to the way the `TarfilePublisherHandler` works:

1. `TarfilePublisherHandler` now accepts an argument `fixed_server` that determines whether or not it should switch servers between retries during a `pubapi_operation`.  The default value of this new argument is `False`, to keep in line with the old behavior.
2. There are new private and public methods to allow callers of `TarfilePublisherHandler` to switch fixed_server behavior on and off.
3. The function `tarfile_in_dropbox` uses (1) and (2) to make sure that the first `exists` call and subsequent `publish` call (if needed) go to the same RCDS server.  It then turns on server switching behavior for all subsequent `exists` calls.
4. The first time an RCDS operation fails, the `TarfilePublisherHandler` retries immediately, and THEN holds off for `RETRY_INTERVAL_SEC`.
5. The `publish` method of `TarfilePublisherHandler` will always switch servers for each retry, regardless of (1) or (2).  When the publish is finished, the `TarfilePublisherHandler` will restore its previous behavior, whether that was server switching or fixed server.
6. Fixed a minor bug with regard to server switching.  The `TarfilePublisherHandler` now keeps track of which server it last contacted, and if server switching is on (the default), then we make sure to actually switch servers each time.

(3) and (4) above are the main suggestions that @DrDaveD, @hgreenlee, and @vitodb raised to me at various points, and are the main drivers of #468 and this PR.

### DIFF
--- a/lib/tarfiles.py
+++ b/lib/tarfiles.py
@@ -488,6 +488,12 @@ class TarfilePublisherHandler:
                     self._fixed_server and always_switch_servers
                 )
 
+                if should_change_selector_behavior and self.verbose:
+                    print(
+                        "TarfilePublisherHandler is configured to use a fixed RCDS server, however this is being overridden "
+                        "for the duration of the current request."
+                    )
+
                 _retry_interval_sec = 0  # First retry immediately, and then we can wait the retry interval
                 # pylint: disable-next=protected-access
                 retry_count = itertools.count(1)
@@ -538,6 +544,10 @@ class TarfilePublisherHandler:
                         _retry_interval_sec = RETRY_INTERVAL_SEC
                     else:
                         if should_change_selector_behavior:
+                            if self.verbose:
+                                print(
+                                    "Restoring fixed server behavior of TarfilePublisherHandler"
+                                )
                             self.__restore_fixed_server_behavior()
                         break
                 return response
--- a/lib/tarfiles.py
+++ b/lib/tarfiles.py
@@ -400,12 +400,15 @@ def tarfile_in_dropbox(args: argparse.Namespace, origtfn: str) -> Optional[str]:
 
 # pylint: disable=too-many-instance-attributes
 class TarfilePublisherHandler:
-    """Handler to publish tarballs via HTTP to RCDS (or future dropbox server)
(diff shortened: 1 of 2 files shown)


shreyb:
All unit tests and checks pass, along with the checks detailed in #468.

shreyb:
Note to reviewers:  There's a big chunk of changes in there (the `pubapi_operation` bit) that's just whitespace, so I'd recommend hiding whitespace changes for this PR.

marcmengel (approved):
Looks really good!

PR #521, opened 2024-01-24
Check if we have enough disk space at strategic points

### LINKED ISSUE #506: Confusing error if $HOME is full.
At the FIFE meeting today it was reported that jobsub_lite gives an unhelpful message if $HOME is out of space/quota. 

Need to figure out how to test that, and improve the error handling.

marcmengel:
This is intended to fix #506, by checking with df / quota if we have enough space to:
* generate rendered condor submit files in submit directory
* copy/change permissions in tarfiles
* copy in executable to submit directory
and directly fail saying there isn't going to be enough space before actually running out.

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -43,6 +43,7 @@ DEFAULT_SINGULARITY_IMAGE = (
 
 
 def cleandir(d: str, verbose: int) -> None:
+    """clean directory out"""
     if not os.path.exists(d):
         return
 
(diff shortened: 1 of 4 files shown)


lib/utils.py
```diff
@@ -485,6 +500,145 @@ def get_client_dn(proxy: Union[None, str] = None) -> Union[str, Any]:
...
+
+def check_space_df(
+    path: str, min_kblocks: int = 100, min_files: int = 20, verbose: int = 0
+) -> Tuple[str, bool]:
+    """check for enough disk space at path, also return fs mountpoint"""
+    cmd = f"df -k {path}"
+    # sample out:
+    # ====================
+    # Filesystem                      1K-blocks        Used Available Use% Mounted on
+    # if-nas-0.fnal.gov:/lbnewc/app 21474836480 20932591360 542245120  98% /dune/app
```
shreyb, line 532: I hate to ask this, but my understanding of the current DOE policy is that we should try to avoid having hostnames out on the public web.  Could you please swap in <HOSTNAME> or something like that in there?

lib/utils.py
```diff
@@ -485,6 +500,145 @@ def get_client_dn(proxy: Union[None, str] = None) -> Union[str, Any]:
...
+    with os.popen(cmd) as pfd:
+        out = pfd.read()
+
+    if out == "":
+        # no quota data...
+        return True
+
+    # sample output
+    # ===================
+    # Disk quotas for user mengel (uid 1733):
```
shreyb, line 589: Same comment as above about hostnames.

shreyb (changes requested):
Looks mostly good.  Two comments:

1.  Please anonymize your commented lines that have hostnames in them (all in utils.py)
2. Just in case some of the systems people decide to submit from either have weird version of `quota` or `df`, or don't have them at all (I think they SHOULD, but I'm not sure about `quota`), I think we should have a flag where the user can opt-out of this. If you agree, I can open a new PR for that after this PR is merged.

shreyb (approved):
LGTM

PR #522, opened 2024-01-25
integration test for disk space 

marcmengel:
Now we verify we get a suitable message with:
* a completely full home area
* one with 3k free (can make stage directory, cannot render files or convert med tarfiles)
* one with 8k free (not enough room for various tarfile things)

Unified the spelling on some of the errors to make the test easier...


marcmengel:
Note: this is going to require merging / rebasing after #522....

lib/tarfiles.py
```diff
@@ -145,6 +145,22 @@ def tar_up(directory: str, excludes: str, file: str = ".") -> str:
 
     tarfile = os.path.basename(f"{directory}{os.getpid()}.tgz")
     check_we_can_write()
+
+    # check for space for tarfile
```
shreyb, line 149: Is the idea here that we don't allow for the creation of the tarfile even if a user says to skip the disk-space check?  Or did you think this check should be included as one that can be skipped with the `--skip-check disk_space` flag/argument?

I'm actually OK with that, since 1) the error message is good and 2) If they're requesting to upload a tarfile, we can assume it's part of the due diligence to check for space even if the user says not to.

shreyb (commented):
I just have a clarifying question.  The code itself looks good to me.

lib/tarfiles.py
marcmengel: No, it was just a race condition:  I was adding a third check for space that I realized I missed in my integration testing, at the same time you were changing the existing tests to be skip-able .  I can go back and add the skip check.

shreyb: Ah got it.  It's up to you - I guess it's a question of whether we want `--skip-check disk_space` to skip _all_ of the checks, or if we want to say, "no, for THIS check, we need to check every time because if it fails, we know the tarfile creation will fail."  Which do you think is the better way?  I'm completely undecided at the moment, because I can see it both ways.

marcmengel: I think this is just as skip-worthy as the others -- That is, I think we *shoudn't* skip any of them, but if we're trying to shave every last tenth of  a second off of submit time (which seems to be the point of the skips), then skipping this one is our biggest win, 'cause we have to do a du -s of the whole area we're tar-ing up to estimate how big the tarfile will be.

marcmengel: So we should skip this third check as well now, if a skip is requested.

shreyb: Okay - you convinced me.  This PR looks good.

shreyb (approved):

PR #523, opened 2024-01-25
Added skip-check option called disk_space to skip disk quota checking

shreyb:
Adds onto #521.  This PR introduces a new option to the `--skip-check` flag, called `disk_space`, that allows a user to run:

`jobsub_submit -G <group> --skip-check disk_space file:///path/to/executable`

That will disable the quota checks implemented in #521.  The default value for this setting is `False`, so the quota checks run by default.

This flag can also be run in combination with `--skip-check rcds`, so a user can skip either or both (or neither) of those checks.  I updated the helptext for `--skip-check` to reflect this.


marcmengel (approved):
This looks good.

marcmengel:
Going to merge this , and but then have to fix #522 for conflicts...

PR #536, opened 2024-02-20
Use spack in wrappers

marcmengel:
This change uses the /cvmfs/fermilab.opensciencegrid.org/packages/common  Spack area to find ifdhc, via a spack environment, 
rather than using the UPS pacakges in /cvmfs/fermilab.opensciencegrid.org/products/common, or looking in /grid/fermiapp
which hasn't actually worked for some time.


templates/simple/simple.sh
```diff
@@ -167,20 +167,9 @@ cat << '_HEREDOC_' > ${JSB_TMP}/ifdh.sh
...
-             setup ifdhc $IFDH_VERSION
-             break
-         else
-            unset PRODUCTS
-         fi
-     fi
-   done
+    . /cvmfs/fermilab.opensciencegrid.org/packages/common/setup-env.sh
+    sos=$(spack arch --operating-system)
+    spack env activate jobsub_env_${sos}_${IFDH_VERSION:-current}
```
shreyb, line 172: A few questions:

1. Is this environment already created?  
2. Do we need to bring whatever configuration is behind that environment into the jobsub_lite repo somehow?
3. Am I reading line 172 correct that if $IFDH_VERSION is not set, then the environment to be activated will be something like `jobsub_env_scientific7_current`?  Do we have an AL9 environment ready to go too?

shreyb (commented):
I just had some questions.  The change itself looks good.

templates/simple/simple.sh
marcmengel: dunegpvm02>$ . /cvmfs/fermilab.opensciencegrid.org/packages/common/setup-env.sh

<dunegpvm02>$ spack env list
==> 6 environments
    jobsub_env_almalinux9_current  jobsub_env_scientific7_current
    jobsub_env_almalinux9_v2_6_20  jobsub_env_scientific7_v2_6_20
    jobsub_env_almalinux9_v2_7     jobsub_env_scientific7_v2_7

The environments in question just have ifdhc and dependencies (mainly python).
Note that the `*_current` environments are just symlinks to the particular versions.

And as usual, these are active only in the subshell of the ifdh.sh script, so as not to mess with any experiment job environments.

shreyb: Awesome - thanks for clearing that up for me.  So you think that we don't need to version-control that here?

I'll wait until tomorrow to merge this, since I'd rather not do it while we're deploying 1.6 in case of a hypotethical emergency bugfix we'd need to do.

marcmengel: Yes.
As to environments,  I can add new ones as part of releasing ifdhc, and we can move the _current links when suitably announced etc. as we do now when making ifhdc current in UPS.   I mean, I could add a script to the jobsub repository that creates the environments for a given version?   It's just:
<pre>
ename=jobsub_env_${sos}_${version}
spack env create $ename
spack -e $ename add ifdhc/version-hash
spack -e $ename install
</pre>
for each new spack instance...

shreyb: Marc and I discussed this offline - mostly about responsibilities (jobsub vs IFDH).  We decided on the following:

1.  The spack environment name will be `ifdh_env_${sos}_${version}` or `ifdh_env_${sos}_current`.  This is so that (a) The name implies that the spack environment is maintained by the IFDHC project, not by the jobsub project; and (b) Marc anticipates that users can be instructed to use that spack environment to select versions of IFDHC, while testing, for example.
2. When the IFDHC developers create a new release of IFDHC, it will be their responsibility to create a new spack environment and to change the `current` environment to point to the new version, on their timeline.
3. This way, when a new IFDHC has been released _for testing_, jobsub will continue to use the old spack environment (still marked `current`) by default.  When the IFDHC developers are satisfied with the test release, then they will change the `current` spack environment to use that new release.

As far as this PR is concerned, two things need to happen:
1. Change line 172 to be `spack env activate ifdh_env_${sos}_${IFDH_VERSION:-current}`
2. Change the current spack `jobsub_env*` environments to be `ifdh_env_*`

After those two are done, we can merge this PR.

shreyb (changes requested):
Marc and I discussed changes that should happen before this PR is merged.  Those are in this latest review.

shreyb:
I just tried running all three of those commands on dunegpvm01, and it didn't work:

```
-bash-4.2$ . /cvmfs/fermilab.opensciencegrid.org/packages/common/setup-env.sh
-bash-4.2$ spack env list
==> 6 environments
    ifdh_env_almalinux9_current  ifdh_env_almalinux9_v2_6_20  ifdh_env_almalinux9_v2_7  ifdh_env_scientific7_current  ifdh_env_scientific7_v2_6_20  ifdh_env_scientific7_v2_7
-bash-4.2$ spack env activate ifdh_env_scientific7_current
==> Warning: could not load runtime environment due to RuntimeError: Unable to locate python command in /cvmfs/fermilab.opensciencegrid.org/packages/common/spack/v0.21.0-fermi/NULL/var/spack/environments/ifdh_env_scientific7_current/.spack-env/view/./bin
```

Is python missing from those ifdh environments?

marcmengel:
Sigh.  I just renamed the environment directories,  but there was a symlink in there that still had the old path...


--
Marc W. Mengel
Computer Services Senior Developer

Data Management and Applications
Fermi National Accelerator Laboratory
630 840 8256 office
www.fnal.gov<https://email.fnal.gov/owa/redir.aspx?C=JkVsWjSyHEuWV2kFJWNPyC9OyMKf7dEIp_yxPk9yfhfadE3mdLfI6T8Ztf8BVaJjzoIe8oJ5H0M.&URL=http%3a%2f%2fwww.fnal.gov>
<https://email.fnal.gov/owa/redir.aspx?C=JkVsWjSyHEuWV2kFJWNPyC9OyMKf7dEIp_yxPk9yfhfadE3mdLfI6T8Ztf8BVaJjzoIe8oJ5H0M.&URL=http%3a%2f%2fwww.fnal.gov>
Connect with us!
Newsletter<https://email.fnal.gov/owa/redir.aspx?C=-29AtpIe40SAvGF4aiNr86KvVjap7dEIuxIdg8Q5bYa9sT7QTFVhxhVnl8zwXkEuBSuAQLRkEfg.&URL=http%3a%2f%2fwww.fnal.gov%2fpub%2ftoday> | Facebook<https://email.fnal.gov/owa/UrlBlockedError.aspx> | Twitter<https://email.fnal.gov/owa/redir.aspx?C=-29AtpIe40SAvGF4aiNr86KvVjap7dEIuxIdg8Q5bYa9sT7QTFVhxhVnl8zwXkEuBSuAQLRkEfg.&URL=https%3a%2f%2ftwitter.com%2fFermilabToday>

shreyb:
No worries - it looks good to me!  I'll go ahead and squash-merge this then.

shreyb:
Actually - doing two quick test submissions to make sure this works onsite and offsite:

8439421.0@jobsub04.fnal.gov  (Onsite)
53143895.0@jobsub03.fnal.gov (Offsite)

shreyb:
Both tests worked.  This is good to go.

shreyb (approved):

Issue #539, opened 2024-02-26
If a disk quota check fails, we shouldn't fail the submission


shreyb:
In 1.6, we added a disk space check to jobsub_lite.  Right now, if that check fails (that is, there's any error in CHECKING the quota), jobsub_lite errors out (see ServiceNow incident INC000001168697).   This was the traceback from that incident:

```
Traceback (most recent call last):
  File "/opt/jobsub_lite/bin/jobsub_submit", line 224, in <module>
    main()
  File "/opt/jobsub_lite/lib/tracing.py", line 172, in wrapper
    res = func(*args, **kwargs)
  File "/opt/jobsub_lite/bin/jobsub_submit", line 189, in main
    set_extras_n_fix_units(varg, schedd_name, cred_set)
  File "/opt/jobsub_lite/lib/utils.py", line 261, in set_extras_n_fix_units
    if not check_space(args["submitdir"], need_blocks):
  File "/opt/jobsub_lite/lib/utils.py", line 512, in check_space
    fs, d_ok = check_space_df(path, min_kblocks, min_files, verbose)
  File "/opt/jobsub_lite/lib/utils.py", line 544, in check_space_df
    availcol = headers.index("Available")
ValueError: 'Available' is not in list
```

I think this is a bug in the behavior, as the quota check could fail due to a number of reasons (like the user submitting it from the wrong dir) that might not actually fail in the jobs getting submitted.

I propose that if the running of the quota check errors out in this way (NOT if the check itself runs properly and returns that there is insufficient space), we should catch that error and print a warning that there was an error running the check, but that we'll try to submit anyway.

shreyb:
@marcmengel What do you think about this idea?  If you agree to it, I don't think we need to bring it to the whole working group, since IMO it's really a bug and not a new feature request.

marcmengel:
This is clearly a bug, or possibly 2 or 3 convolved :-(.    If I'm parsing the output, I should be setting
LC_ALL=C or some such and I'm not, second, even so there could be some reason we don't 
get any output from the command, and we should just admit we can't check here and ignore it...

retzkek:
So I want to back up here. Why was the decision made to solve #506 by checking disk space and quota instead of improving the error handling of os errors? We should be looking to simplify whenever possible, not add more complexity. It’s supposed to be “lite” right?

marcmengel:
Checking disk space first is better, because filling up the disk and then noticing you've
[...]

PR #541, opened 2024-02-28
Fix 539

marcmengel:
This is a first fix of #539.


marcmengel:
CLosing this out, we decided to improve the error handling elsewhere and not try to do df / quota checks.

PR #546, opened 2024-03-04
Take direct quota checks out of code, and try to improve error handling for out of space errors

### LINKED ISSUE #539: If a disk quota check fails, we shouldn't fail the submission

shreyb:
Closes #539 by removing quota checks altogether, as requested by @retzkek  in #539 discussion.

I'd like to note that while this may address the issue immediately, a few issues in the testing came up by the fact that we run commands using `os.system` rather than the `subprocess` functions.  I'm going to open a new issue specifically to address that in the `tarfiles.tar_up` func, since that's the one that made things tougher here.

Another note:  For `render_files`, I had to split the `render_files` func up to make our fake filesystem tests work with functions that have tracing decorators.  The trace should still look the same, and we may want to do for other functions as well.

Final note:  The fake filesystem was implemented using (formerly Google's) `pyfakefs` pytest plugin.  To run those unit tests, the reviewer will have to run `pip install pyfakefs`.


marcmengel (approved):
Looks good to me.  It does remind me that the GZIP=-n in the tarfile creation is now generating a warning on AL9, but that is another issue (soon to be an Issue in the tracker).

PR #551, opened 2024-03-13
Make sure --global-pool doesn't clear the pre-existing environment

shreyb:
This fixes the bug brought up in the discussions offline.

It also allows us to pass verbose mode to the underlying condor_vault_storer call made by condor_submit, either through the environment or via the verbose flag.

I ran all the unit and integration tests to validate these changes.

### DIFF
--- a/lib/condor.py
+++ b/lib/condor.py
@@ -295,8 +295,13 @@ def submit(
     # the updated one in the condor release
     #
     jldir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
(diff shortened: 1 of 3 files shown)


shreyb:
@DrDaveD :  Included in this PR is the ability for us to pass verbose or debug flags to condor_vault_storer by doing something like:

`_condor_SEC_CREDENTIAL_STORER="/path/to/condor_vault_storer -d" jobsub_submit <args>`.

More directly, the `jobsub_submit` `--verbose` flag will also trigger `-v` being sent to `condor_vault_storer`.

marcmengel (approved):

PR #557, opened 2024-03-31
This fixes a bug introduced in jobsub_lite 1.7

shreyb:
In #551, we changed the behavior of how the `_condor_SEC_CREDENTIAL_STORER` environment variable was used.  Previously, the `condor.py:submit()` function would always coerce that environment variable to point to the jobsub_lite version of `condor_vault_storer`.  #551 changed that so that if `_condor_SEC_CREDENTIAL_STORER` was defined prior to submission, that setting was respected.  This broke POMS submissions, as POMS relies on jobsub's overriding whatever setting for `_condor_SEC_CREDENTIAL_STORER` that exists.

Originally, that portion of #551 was put in place to allow us to pass verbose and debug to the underlying `condor_vault_storer` commands for troubleshooting purposes.  This PR keeps that behavior by using the user-passed `--verbose/--debug` flags while restoring the previous overriding of `_condor_SEC_CREDENTIAL_STORER`.

I also added tests to make sure we didn't accidentally re-break this behavior in the future.

### DIFF
--- a/lib/condor.py
+++ b/lib/condor.py
@@ -296,13 +296,17 @@ def submit(
     #
     jldir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
-    _sec_cred_storer_val = os.environ.get("_condor_SEC_CREDENTIAL_STORER", None)
-    if _sec_cred_storer_val is None:
-        _sec_cred_storer_val = (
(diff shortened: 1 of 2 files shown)


marcmengel (approved):
Looks good to me.

PR #562, opened 2024-04-05
Stealing token compat from ifdh.cfg for fake_ifdh.py

### LINKED ISSUE #561: Log web sandbox creation works only with proxy
When submitting jobs with token only using:
```
jobsub_submit --debug -G fermilab --memory 500MB --disk 500MB --auth-methods token file:///usr/bin/sleep 300
```
in the submission log we see
```
[...]
Transferring files to web sandbox...
gfal-mkdir error: 13 (Permission denied) - HTTP 401 : Authentication Error 
warning: error creating sandbox, web logs will not be available for this submission: Error: Unable to make directory https://fndcadoor.fnal.gov:2880/fermigrid/jobsub/jobs/2024_04_04/4111a9ea-f1e1-432d-a4f3-31109f3fee22
[...]
Checking if fermilab_5c91474037 -S ",compute.read,compute.create,compute.cancel,compute.modify,storage.read:/fermilab/users/vito,storage.create:/fermilab/users/vito,storage.create:/fermigrid/jobsub/jobs," credentials exist
[...]
```
the token scope is supposed to have write access for the web sandbox `storage.create:/fermigrid/jobsub/jobs`
so the error seems to imply the code creating the web sandbox is using proxy authentication.
As result the job runs and is able to upload its log, I guess this time using tokens for auth,
but in the web sandbox we are missing:
```
simple.cmd
simple.sh
sleep
```
If the submission doesn't use `--auth-methods token` all is working as expected, the web sandbox is created at job submission time and all expected files are present in there.

marcmengel:
Token-only launches were failing to copy sandbox files at launch... Needed 
BEARER_TOKEN set in the environment for gfal utilities.  Fixes #561

### DIFF
--- a/lib/fake_ifdh.py
+++ b/lib/fake_ifdh.py
@@ -385,9 +385,7 @@ def getProxy(
 
 
 # pylint: disable=invalid-name
-gfal_clean_env = (
-    "unset PYTHONHOME PYTHONPATH LD_LIBRARY_PATH GFAL_PLUGIN_DIR GFAL_CONFIG_DIR"
-)
+gfal_clean_env = "unset PYTHONHOME PYTHONPATH LD_LIBRARY_PATH GFAL_PLUGIN_DIR GFAL_CONFIG_DIR;  [ -s ${BEARER_TOKEN_FILE:-/dev/null} ] && export BEARER_TOKEN=`cat ${BEARER_TOKEN_FILE}`"
 
 
 def fix_pnfs(path: str) -> str:


shreyb (changes requested):
Looks good, but two things.

1.  Could you please update the submodule commit hash?
2. I think the answer is yes to this question, but just to be sure - by putting this new setting in `gfal_clean_env`, we're making sure it only gets used in the various gfal commands here in fake_ifdh, correct?

marcmengel:
RIght; so that only happens in the subshell we run the gfal-copy in; we scrub the various PYTHON related variables (because the gfal developers don't know about 
"python -I " ?) and grab BEARER_TOKEN from  BEARER_TOKEN_FILE, if it is set.

shreyb (approved):
Cool - I just did a quick test, and this worked as advertised.  Merging.

Issue #544, opened 2024-02-29
New condor_vault_storer version


DrDaveD:
Please include the `condor_vault_storer` from [HTCONDOR-2232](https://github.com/htcondor/htcondor/pull/2232) in a jobsub_lite release.
Here's a [direct link](https://raw.githubusercontent.com/htcondor/htcondor/32a5b5794979fc439323809bbfa5b4f0e1890f6b/src/condor_credd/condor_credmon_oauth/condor_vault_storer).

DrDaveD:
There was now also an additional change from another PR, both of which are now merged and scheduled for the 23.0.8 condor release.  Use instead [this link](https://raw.githubusercontent.com/htcondor/htcondor/V23_0-branch/src/condor_credd/condor_credmon_oauth/condor_vault_storer).

DrDaveD:
Note that the way to enable verbose mode changes in this new version of condor_vault_storer, because the old way doesn't work with condor 23.  So the method added in #551 should change to setting `_condor_SEC_CREDENTIAL_VAULT_STORER_OPTS` instead of appending to `_condor_SEC_CREDENTIAL_STORER` at the same time as this is issue is addressed.

shreyb:
Thanks for drawing our attention to that, Dave.

We discussed this at the jobsub meeting, and the plan is to do this entire transition in phases:

1) Do what Dave requested here.  This issue will become the first item in a new project I'll open for jobsub_lite.  Target this for release 1.8, due out in April or May.
2) Sometime between now and June 30, 2024, all of the interactive nodes will be upgraded to AL9.  At that point, condor will be upgraded to at least 2023.0
3) Once [HTCONDOR-2232](https://github.com/htcondor/htcondor/pull/2232) is in a condor release (probably late Summer to early Fall of 2024), upgrade interactive nodes to that version.
4) Wait until [POMS-16](https://github.com/fermitools/poms/issues/16) is finished.
5) Put out a test release where we remove condor_vault_storer from the jobsub repository.
6) Once we are satisfied with testing, put out a final release with condor_vault_storer removed from the jobsub repository.  At that point, we will expect condor_vault_storer to come from condor.

PR #563, opened 2024-04-11
Updating condor_vault_storer

marcmengel:
Latest from @DrDaveD .   Implements #544 .

### DIFF
--- a/bin/condor_vault_storer
+++ b/bin/condor_vault_storer
@@ -4,13 +4,10 @@
 # Can be run automatically from condor_submit when configured as
 #   SEC_CREDENTIAL_STORER.
 
-# debugging -- turn on by uncommenting
-#lf=/tmp/csv_$$.log
-#exec 2> $lf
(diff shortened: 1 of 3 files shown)


marcmengel:
Still waiting for integration tests to complete...

marcmengel:
Okay, integration tests look good.   There is a problem with test_condor_submit_dag1() in the integration tests (start job keeps going held due to system periodic hold) , but it also happens in the master branch without these changes, so I don't think this causes it.

shreyb:
Note that the issue @marcmengel is referring to he fixes in #564. (Marc, please correct me if I'm mistaken about that).

shreyb (approved):
Looks good.  I ran the unit tests affected, and everything works.

Also, doing a diff directly on this PR's version of condor_vault_storer and the one in @DrDaveD's link in #544 , shows just the POMS clause, which is what we expect.

PR #564, opened 2024-04-12
fix condor_submit_dag test

marcmengel:
It needed some environment variables passed; not sure how they were getting through before, but they aren't lately (alma9 container? condor update?) .  Anyhow, I think the test works again now.


shreyb:
LGTM, but one question.  Do we know that this fixes the test?

marcmengel:
So I thought I answered this; yes this does fix it.

shreyb (approved):
Great - the test worked for me.

PR #565, opened 2024-04-15
cleanup copied tokens

marcmengel:
we make a copy of our token because the submission monkeys with it, but
we weren't cleaning it up...

### DIFF
--- a/lib/utils.py
+++ b/lib/utils.py
@@ -31,6 +31,7 @@ import uuid
 
 import classad  # type: ignore # pylint: disable=import-error
 from tracing import get_propagator_carrier
+import token_mods
 
 from creds import CredentialSet
 import version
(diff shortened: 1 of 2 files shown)


lib/utils.py
```diff
@@ -68,6 +69,11 @@ def cleanup(varg: Dict[str, Any]) -> None:
             if entry.name.startswith("js_") and time.time() - sb.st_mtime > 604800:
                 cleandir(entry.name, verbose=varg["verbose"])
 
+    # if our BEARER_TOKEN_FILE ends with our pid, then we copied it and should
+    # clean it up...
+    if token_mods.is_copied_token(os.environ["BEARER_TOKEN_FILE"]):
```
shreyb, line 74: Can we also add a verbose clause here, where if `varg["verbose"] > 0`, we print that we cleaned up that file?

shreyb (changes requested):
Looks good.  I have one tiny request.

lib/utils.py
marcmengel: Done.

shreyb: Did you push that commit to Github?  The PR looks the same as before.

marcmengel:
Sorry, @shreyb  I *thought* I pushed it...  too many balls in the air this week... sigh.

PR #567, opened 2024-04-19
clear up escape warning in tests

marcmengel:
Some strings in the test should have had r-prefixes.


shreyb (approved):
Looks good.  Thanks for cleaning that up.

PR #568, opened 2024-04-20
When --global-pool flag is passed, make sure we save the BEARER_TOKEN_FILE and X509_USER_PROXY into the restored environment post-submission

shreyb:
This fixes a bug that was introduced with the combination of #565 and #551.  Namely, for the `--global-pool` case, we were saving the environment before modifying the collector host, and then directly before submission, restoring it with `packages.orig_env()`.  

In this bug, since the original environment is saved during argument parsing, the credentials have not yet been gotten, and thus the `BEARER_TOKEN_FILE` and `X509_USER_PROXY` environment settings were wiped when `packages.orig_env()` is called in `condor.submit()`.

This PR fixes that bug, revealed in the testing of the 1.8-rc1 release.

### DIFF
--- a/lib/creds.py
+++ b/lib/creds.py
@@ -18,6 +18,7 @@ import os
 from typing import Any, Dict, Optional, List
 
 import fake_ifdh
+import packages
 from tracing import as_span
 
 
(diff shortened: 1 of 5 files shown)


shreyb:
Note:  With this fix, all unit tests now pass in AL9.

marcmengel (approved):

PR #572, opened 2024-05-17
If cigetcert fails for any reason, raise `PermissionError`

### LINKED ISSUE #570: Fail getting proxy if cigetcert fails for any reason
The check only actually returns an exception for one specific error condition, continuing on for others, which still may not have gotten a certificate. This can result in confusing errors like `PermissionError: Failed attempting 'voms-proxy-init ...`

https://github.com/fermitools/jobsub_lite/blob/115727dd2d354331dbbd87c9b93c82eb69bb9e37/lib/fake_ifdh.py#L360-L365

shreyb:
Fixes #570.

This PR also adds two new unit tests to check the new logic, as well as a minor bug fix for the existing getProxy tests themselves.

Note:  According to the docs:

- https://docs.python.org/3.7/library/subprocess.html#subprocess.run
- https://docs.python.org/3.7/library/subprocess.html#subprocess.CompletedProcess.check_returncode

there are two ways to check if a `subprocess.run()` call fails:

1.  Pass the `check=True` kwarg to the function, which will make it raise a `CalledProcessError` if the return code of the underlying subprocess is nonzero (i.e. `subprocess.run('exit 1', check=True)` will raise a `CalledProcessError`), and
2. If you don't want to immediately raise the error, set `check=False`, and then save the return value of `subprocess.run()` to a variable (e.g. `result`).  Then, when you're ready, you can run `result.check_returncode()`, which will raise the `CalledProcessError` if the return code is nonzero.

We use the second method here, since we want to raise different errors based on the output of `cigetcert`.

### DIFF
--- a/lib/fake_ifdh.py
+++ b/lib/fake_ifdh.py
@@ -348,21 +348,26 @@ def getProxy(
 
     if force_proxy or invalid_proxy:
         cigetcert_cmd_str = f"cigetcert -i 'Fermi National Accelerator Laboratory' -n --proxyhours 168 --minhours 167 -o {certfile}"
-        # pylint: disable=subprocess-run-check
-        # TODO:  See if we can put check in here, catch the error, and raise our Exception as below # pylint: disable=fixme
         cigetcert_cmd = subprocess.run(
             shlex.split(cigetcert_cmd_str),
             stdout=subprocess.PIPE,
             stderr=subprocess.STDOUT,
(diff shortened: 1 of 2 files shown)


goodenou:
These changes look good Shreyas.  Good ahead and merge the PR.

shreyb:
Merging, as Lisa gave her written approval in the comment above.

Issue #574, opened 2024-09-12
CVMFS logs in job logs are not properly retrieved


vitodb:
In jobsub stderr logs CVMFS logs are not properly retrieved.
For example in this [job stderr log](https://landscape.fnal.gov/lens/view/job/13295461.3375@jobsub05.fnal.gov/stderr) toward the bottom there is:
```
cvmfs info:
sed: -e expression #1, char 31: unterminated address regex
```

marcmengel:
This would almost certainly be the sed filter line set at 
https://github.com/fermitools/jobsub_lite/blob/master/templates/simple/simple.sh#L455

so one possibility I see right away is in the simple.sh template
https://github.com/fermitools/jobsub_lite/blob/master/templates/simple/simple.sh#L443
we have a "dune" where we ought to have a {{group}} when trying to get the second latest log message,
So this might work for "dune" but not others(?).

vitodb:
Other than replacing "dune" with  {{group}}  as mentioned by Marc, I think we need another change in this line 
https://github.com/fermitools/jobsub_lite/blob/master/templates/simple/simple.sh#L443
here for the code
```
second_latest_rev_re=$( echo "$dummyline"; attr -g logbuffer /cvmfs/dune.opensciencegrid.org/ | 
```
only the `attr` output is passed through the pipe and folloing chained commands, this is causing the variable `second_latest_rev_re` to be on two lines.
Later on this variable is used in 
https://github.com/fermitools/jobsub_lite/blob/master/templates/simple/simple.sh#L455
and being `second_latest_rev_re` on two lines is breaking the `sed` expression.

I think line 443 should be replaced by:
```
second_latest_rev_re=$( (echo "$dummyline"; attr -g logbuffer /cvmfs/dune.opensciencegrid.org/) | 
```
to let the "echo" output to go through the pipe and the other chained commands.

PR #581, opened 2024-10-04
Update cvmfs_info function

vitodb:
Address issue #574
Extend the use of cvmfs_info function to osgstorage repos alongside opensciencegrid repos
For SBND and ICARUS check also SBN CVMFS repos

### DIFF
--- a/templates/simple/simple.sh
+++ b/templates/simple/simple.sh
@@ -39,12 +39,10 @@ fi
 
 # Set up parameters used by cvmfs_info function
 CVMFS_REPO_TYPE_LIST=(opensciencegrid osgstorage)
-{%if group in ('sbnd','icarus') %}
-CVMFS_REPO_LIST=({{group}} sbn)
-{% else %}
-CVMFS_REPO_LIST={{group}}
-{% endif %}
-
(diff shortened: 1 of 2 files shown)


marcmengel (approved):
Looks good.

marcmengel:
This is a good idea ; should we also, for suitable experiments, be checking larsoft.opensciencegrid.org as well?

vitodb:
> This is a good idea ; should we also, for suitable experiments, be checking larsoft.opensciencegrid.org as well?

I thought of this, but because for larsoft CVMFS repo we don't have the osgstorage repo I left this over.
However adding an extra check to exclude osgstorage, we can test also larsoft for suitable experiments.

vitodb:
I have included the check of larsoft CVMFS repo as suggested by Marc, alongside few other improvements.

shreyb:
@marcmengel I'll hold off on merging for a couple of days in case you wanted to check Vito's latest commit.

marcmengel (commented):
Looks good!  I like the conditional larsoft check...

marcmengel (approved):

PR #582, opened 2024-10-07
Remove all /grid/fermiapp references or update them to point to /cvmfs/fermilab.opensciencegrid.org

### LINKED ISSUE #580: Update old defunct BLUEARC to new ceph mount in templates and tests
The old BLUEARC mount used for various source scripts (for tests, IFDH, etc.) is now defunct and is being decommissioned.  We need to upgrade those paths to the new CEPH mount at Fermilab.  The areas concerned are tests and wrapper scripts, as far as I can see.

shreyb:
As the title says.

Closes #580


vitodb (approved):
Looks good!
I see there are few `/fnal/ups/etc/setups.sh` occurrences, I'm wondering if we want to remove these as well.

PR #583, opened 2024-10-15
Fix RUNTIME for running jobs

### LINKED ISSUE #501: RUNTIME mismatch
jobsub_lite uses the `RemoteWallClockTime` classad for a job to get its RUNTIME attribute for running jobs in `jobsub_q` output.  A user rightly noted that this doesn't match the output of `condor_q`.  This seems to be because condor_q, when run, is querying its own internal database, and may NOT be updating the classad:

https://github.com/htcondor/htcondor/blob/ae1a4a42d6e4f8994575a67ff5217e131e713f6c/src/condor_q.V6/queue.cpp#L4226

We should ask the condor folks about this and how to get an accurate RUNTIME.

shreyb:
Closes #501 

Sometime in the last few Condor releases, the classad `RemoteWallClockTime` stopped being populated in a job's classad while a job was running, and that broke our `jobsub_q`.  After consultation with the condor developers, this PR has the new implementation of calculating the job's cumulative run time.

The calculation for a running job is:  `(ServerTime - ShadowBday) + RemoteWallClockTime`.

For an idle, completed, or held job, the classad `RemoteWallClockTime` still works, and that's reflected in this PR.

This PR also includes a small tweak to the `jobsub_totals` script to improve performance; and a few comments to make clear what Condor expression corresponds to what column.

### DIFF
--- a/bin/jobsub_cmd
+++ b/bin/jobsub_cmd
@@ -206,11 +206,12 @@ def main() -> None:
                     "%-11s ",
                     'formatTime(QDate,"%m/%d %H:%M")',
                     "-format",  # RUNTIME
-                    # For running jobs, use ServerTime - ShadowBday.  Otherwise, use RemoteWallClockTime
+                    # For running jobs, use ServerTime - ShadowBday, and add it to the accumulated RemoteWallClockTime for previous executions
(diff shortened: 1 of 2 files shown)


marcmengel (approved):
Looks good.

PR #587, opened 2024-10-21
Put line continuation backslash back into fife_wrap tests

shreyb:
I accidentally broke the integration tests that depend on `fife_wrap` with #582.  This PR fixes that bug.


vitodb (approved):
looks good!

PR #588, opened 2024-10-25
Add --gpu flag to submit jobs to GPU resources

### LINKED ISSUE #505: Add --gpu flag to jobsub_submit that allows users to request GPUs
Currently, if users want to request GPUs, they use the following incantation:

```
jobsub_submit -G <group> --lines='+RequestGPUs=1' <other_args>
```

We should have a flag, `--gpu`, that accepts a number (defaulting to `1`), which will add the `+RequestGPUs` classad to the JDF.

We might also want to build in another flag where the user can specify the kind of GPU, but that's still under discussion (currently, users can specify that using `--lines`, so there's no hurry).

shreyb:
Closes #505

This new flag sets the following classads in the generated Job Definition File:

* `request_GPUs`  (for HTCondor:  See https://htcondor.readthedocs.io/en/latest/users-manual/submitting-a-job.html#jobs-that-require-gpus)
* `RequestGPUs` (For the OSG/GlideinWMS.  This flag has been in use for Fermilab's GWMS Frontends, and seems to be an alternate convention for jobs running on the OSG).

All unit and integration tests pass with this PR.

### DIFF
--- a/lib/get_parser.py
+++ b/lib/get_parser.py
@@ -305,7 +305,9 @@ def get_parser() -> argparse.ArgumentParser:
         default=os.environ.get("CMTCONFIG", ""),
         help=" Set up minervasoft release built with cmt configuration. default is $CMTCONFIG",
     )
-    parser.add_argument("--cpu", help="request worker nodes have at least NUMBER cpus")
+    parser.add_argument(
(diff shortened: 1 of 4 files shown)


vitodb (approved):
Looks good

PR #589, opened 2024-10-29
Fix jobsub fetchlog verbose

shreyb:
In the course of investigating SNOW ticket INC000001183512, I found that there was a bug in how `jobsub_fetchlog` called `utils.cleanup` in the case that it was pulling logs from the condor schedd.  I fixed that bug here, as well as strengthening the error handling in the underlying `utils.cleanup` function.


vitodb (approved):
This looks good!

cathulhu (approved):
Makes sense, approved

PR #592, opened 2024-12-11
Break submit trace into two spans and add tracing to wrapper 

retzkek:
This breaks the initial trace into two spans: a `jobsub` overarching span and a `jobsub_submit` span that contains the submit steps. This way any subsequent spans are not shown as being part of the submit, and the submit span can be collapsed without hiding them. There's still a bit of discontinuity where the `jobsub` span ends with the submit and does not encapsulate downstream spans, but that would be tough to avoid and I don't think it causes any major issues, the trace otherwise appears OK in Jaeger.

Further, this roughly adds a tracing span to the job wrapper (only `simple.sh` for now). I anticipate some iteration here before merging and including in DAGs. Some key items:
* Span ID generation
* Environment variables for endpoints and options

Screenshot 2024-12-11 at 12-23-02 Lightning talk distributed tracing - Google Slides
Screenshot 2024-12-11 at 12-23-12 Lightning talk distributed tracing - Google Slides
Screenshot 2024-12-11 at 12-30-19 Lightning talk distributed tracing - Google Slides

### DIFF
--- a/templates/simple/simple.sh
+++ b/templates/simple/simple.sh
@@ -135,8 +135,43 @@ redirect_output_finish(){
     {%endif%}
 }
 
+trace_start(){
+    if ( is_set $TRACEPARENT ); then
+        version="00"
+        export JSB_TRACE_ID=`echo $TRACEPARENT | cut -d'-' -f 2`
+        export JSB_PARENT_SPAN_ID=`echo $TRACEPARENT | cut -d'-' -f 3`
+        sample=`echo $TRACEPARENT | cut -d'-' -f 4`
+        # 8-byte random span ID e.g. adce7acee6441ec74
+        export JSB_SPAN_ID=`head -c8 /dev/urandom | hexdump -e '"%02x"'`
+        # epoch nanoseconds e.g. 1544712660000000000
+        export JSB_SPAN_START=`date +%s%N`
+        # create child span
+        export TRACEPARENT="$version-$JSB_TRACE_ID-$JSB_SPAN_ID-$sample"
+    fi
+}
+
+trace_finish(){
+    # TODO check some other env vars
+    if ( is_set $OTEL_EXPORTER_OTLP_ENDPOINT ); then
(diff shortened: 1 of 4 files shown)


shreyb:
Discussion points:

Very good, but we need the following:
1) Error handling in the wrapper script
2) Standardization of ENV VARS that we want to pass to jobs

templates/simple/simple.sh
```diff
@@ -135,8 +135,44 @@ redirect_output_finish(){
...
 }
 
+trace_start(){
+    if ( is_set $TRACEPARENT ); then
+        version="00"
+        export JSB_TRACE_ID=`echo $TRACEPARENT | cut -d'-' -f 2`
+        export JSB_PARENT_SPAN_ID=`echo $TRACEPARENT | cut -d'-' -f 3`
+        sample=`echo $TRACEPARENT | cut -d'-' -f 4`
+        # 8-byte random span ID e.g. adce7acee6441ec74
+        export JSB_SPAN_ID=`head -c8 /dev/urandom | hexdump -e '"%02x"'`
```
retzkek, line 145: Any better portable ideas to generate a span ID?

`uuidgen` was suggested, which appears to be available alongside `hexdump`, but the output would need to be munged a bit - we need 8 bytes/16 chars, so could be the last two parts, which should be unique even if `uuidgen` falls back to time-based (which is probably unlikely). Is that better? ¯\\\_(ツ)\_/¯

```
Apptainer> for x in {1..5}; do uuidgen | tr -d '-' | cut -c '17-'; done
98ea0d9203e32de4
a670a63510d929e5
a362b70ee7c45a2c
adbdcecf9ff8fe00
a003fc48358eb47b
Apptainer> for x in {1..5}; do uuidgen -t | tr -d '-' | cut -c '17-'; done
9604001a4af11e5f
9d58001a4af11e5f
a312001a4af11e5f
a2ed001a4af11e5f
b76a001a4af11e5f
```

PR #594, opened 2024-12-20
Jobsub subcomand parser and API interface

marcmengel:
Subcommand Parser

You can now say "jobsub submit ..." instead of "jobsub_submit ..." and jobsub gives help for all the commands.

There is also a jobsub_api which you use by saying:
 ` jobsub_call( [argv] , collect_ouput) `  Which if you pass collect_ouput = True will return the commands output as a string, otherwise the output goes to stdout.  There are some utility regexes to parse said output;
a simple example follows:

```
from jobsub_api import jobsub_call, jobsub_submit_re, jobsub_q_re

def test_api_demo():
    group = "fermilab"
    out1 = jobsub_call(
        ["jobsub_submit", "-G", group, "file://./job_scripts/lookaround.sh"], True
    )
    m1 = jobsub_submit_re.search(out1)
    if m1:
        jobid = m1.group("jobid")

        time.sleep(1)

        out2 = jobsub_call(["jobsub_q", "-G", group, jobid], True)
        for line in out2.split("\n")[1:]:
            m = jobsub_q_re.search(line)
            if m:
                print(f"saw job {m.group('jobid')} command {m.group('command')}")
    else:
        print("submission failed: output:\n", out1)
```


marcmengel:
So this looks like a HUGE change, because code moved from the various jobsub_xxx scripts into lib/mains.py
and then was sliced up into routines.  But if you snip out the moved code and diff the chunks, it's much more reasonable to look at: here it is:

[compare_moved.txt](https://github.com/user-attachments/files/18206078/compare_moved.txt)

marcmengel:
At this stage, lots of things work, but I need to run and clean up from the whole test suite.  So look
for a few more changes before this is ready for prime time.

shreyb:
Do you want to move the PR into draft status until it passes the tests, or do you think it's close enough at this point?

marcmengel:
I think it's about ready.  The only test I don't see why it's failing is `test_get_schedd_1`, 
Maybe another pair of eyes on that would help.

shreyb:
`test_get_schedd_1` passes when I run it with this PR checked out, actually.  What error are you getting?

I'm not sure why the pylint is now failing though - it looks like import errors, but those worked before.  I'll check that out as well.

bin/jobsub
```diff
@@ -1,64 +1,68 @@
...
-         Print JSON dump of token files
+PREFIX = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
-    All josub_lite commands will print usage with --help, and
-    "man jobsub_..." will give a manual page, or see:
-    https://fifewiki.fnal.gov/wiki/Jobsub_Lite#User_Documentation
+sys.path.append(os.path.join(PREFIX, "lib"))
 
-jobsub_lite also wraps the following condor commands to accept '-G group':
+# pylint: disable=wrong-import-position, ungrouped-imports
```
shreyb, line 16: pylint is complaining because in all the other executables, we ignore `import-error` at this point.  Could you please add `import-error` to this list?  I also believe that you can't have a space between checks to ignore, so the whole line should read:

`# pylint: disable=wrong-import-position,ungrouped-imports,import-error`

This should fix the pylint checks.

shreyb (changes requested):

bin/jobsub
marcmengel: Done.   Its weird though, I have the pre-commit stuff turned on and it was passing pylint for me...  Maybe I have an old version?  Sigh. 
I'm also still seeing test_get_schedd_1 fail, it keeps getting jobsu01 instead of jobsubdevgpvm01 (?!) 
And test_condor_submit_dag1 is failing for me: well the test runs, but its output doesn't show up, so the failure shows up in check_output... 
If I look interactively, jobsub_fetchlog for its job doesn't get any output, but jobsub_fetchlog -condor does; so
something hosey in getting the tokens to the job(?).   I mean, it doesn't run long enough for them to expire or
anything.

marcmengel: Hm... I'm wondering if it is because the .cmd file has a hardcoded, old, `+JobsubOutputURL=` line... perhaps
one of us owns that directory and the other can't put stuff there(?)...   I will see about tweaking that...

shreyb: For the test_get_schedd_1 failure, what condor_status command is being run?

marcmengel: It uses the python library Collector class to do a query... 
`IsJobsubLite=?=true && STRINGLISTIMEMBER("fermilab", SupportedVOList) && regexp(".*dev.*", Machine) && InDownTime != true`
and of course is now working for me if I run it standalone...

shreyb: Sorry Marc - I'm a bit confused.  Am I understanding correctly that if you run the test by itself, it works, but if you run it as a part of the other tests, it fails?

marcmengel: At the time of that writing, yes.  Then it failed on fifeutilgpvm03, but passed on my desktop.  Now it's running clean again on fifeutilgpvm03.   All without changing the code, or as far as I know, any of the other related software... So ... gremlins(?) aliens(?) nyetworking(?).

marcmengel:
Well, I've gotten completely sidetracked here on to this condor_submit test; I'm going to leave it broken for now; its having trouble copying the output files back to where jobsub_fetchlog wants to find them, and *then* it's finding the wrong flavor of ifdhc, and getting lbrary link errors...  I'm going to leave that broken for now and redo the submit files for that test separately.

marcmengel:
Please ignore my last comment; that was before I read the patch :-)

shreyb:
Alright - so should we go ahead and start reviewing this PR with an eye to fixing the condor_submit test before the next release?

marcmengel:
Yes please.  The problems with the condor_submit test are longstanding; it was passing for the wrong reasons previously, and there are several issues to debug so that the job submitted by condor_submit is retrievable by jobsub_fetchlog without reusing the same file output directory, etc.

lib/mains.py
```diff
@@ -0,0 +1,738 @@
...
+
+    # pylint: disable=too-few-public-methods
+
+    def __call__(self, parser, namespace, values, option_string=None):  # type: ignore
+        os.environ["GROUP"] = values
+        setattr(namespace, self.dest, values)
+
+
+def jobsub_cmd_parser(
+    jobsub_q_flag: bool, parser: Optional[argparse.ArgumentParser]
```
shreyb, line 241: Two nitpicky things:

1. Reading through this so far, it seems like `jobsub_cmd_parser` is more concerned with whether or not there's a passed in `parser`, rather than if the command is jobsub_q or not.  As such, would you be able to switch the argument order of this func, so it's `jobsub_cmd_parser(parser, jobsub_q_flag)`?  
2. Should the two arguments have default values?
3. I actually didn't quite get on first read-through what the `jobsub_q_flag` bool was supposed to signify.  Could we change the name to something like `is_cmd_jobsub_q`, `called_by_jobsub_q` or even `is_jobsub_q`?  Something that makes it very clear that the function cares about whether or not the caller is `jobsub_q`?  Alternatively, to make it even cleaner, we could even disassociate `jobsub_q` from this function, and just indicate whether or not our returned parser should include the `--user` option: so maybe the argument should be `need_user_flag`.  Thoughts?

bin/condor_submit_dag
```diff
@@ -59,14 +59,16 @@ def main():
...
+    parser = get_parser.get_submit_parser()
 
     arglist, cmd_args = parser.parse_known_args()
 
     # pull dag file out of args
     if len(cmd_args):
         f = cmd_args[-1]
         cmd_args = cmd_args[:-1]
+    else:
+        raise SystemExit("Filename argument is required.")
```
shreyb, line 71: This exception will not get caught by our general error handling (e.g. in jobsub_submit), since `SystemExit` doesn't inherit from `Exception` (it inherits from `BaseException`).  

1. Is that OK?
2. If not, would you rather change this exception class here (maybe a `RuntimeError`?) , or change all of the general handling clauses to catch `BaseException`?

marcmengel: We could raise argparse.ArgumentError... But this is the mainline for the condor_submit_dag wrapper, and there isn't any exception handling upstream of it.

shreyb: ....you're absolutely right. Please ignore this comment.

config
shreyb, line 1: Could you please update this config reference to point to the proper commit (503e7b)?

lib/mains.py
shreyb, line 1: So I worry about having this huge file with all the various mains because the size of the file just makes it a bit less grok-able.  What would you think about putting these in a subdirectory - so the structure could be something like:
```
lib
|
|- condor.py
|- dagnabbit.py
|...
|- mains
    |- __init__.py
    | - submit.py
    |- cmd.py
    |- fetchlog.py
|- tarfiles.py
...
```

Then the imports throughout would change. but it'd have the benefit of keeping each file's scope pretty-well defined.

Thoughts?

I'm also curious to hear if it's just me - @retzkek,  @vito, or @cathulhu  - any thoughts on this?

lib/jobsub_api.py
```diff
@@ -0,0 +1,62 @@
+from typing import List, Optional, Generator
+import re
+import sys
+import contextlib
+from io import StringIO
+from mains import jobsub_submit_main, jobsub_fetchlog_main, jobsub_cmd_main
+
+
+@contextlib.contextmanager
+def output_saver(should_i: bool) -> Generator[StringIO, StringIO, StringIO]:
```
shreyb, line 10: Perhaps I'm missing something, but shouldn't this type expression be `Generator[StringIO, None, StringIO]`, since we're not sending anything to the generator?

marcmengel: Hm.. well PEP 484 says: "The return type of generator functions can be annotated by the generic type Generator[yield_type, send_type, return_type] provided by typing.py module:"  so I think that means
it should be Generator[StringIO, bool, StringIO]  because we pass a boolean in.

shreyb: Well, that PEP mentions the `send_type`, which I don't believe applies here.  The `send_type` as I understand it is only if you're sending a value to the generator, which we're not - we're instantiating the generator with a bool that affects its behavior.

If we look at https://peps.python.org/pep-0484/#annotating-generator-functions-and-coroutines, it shows an example, where they're sending values to the generator like 

```python
res = yield round(res)
```

In that example, they're passing floats to the generator, which rounds them and then yields the rounded value back to the caller, waiting for the next value to round.

I don't believe we're doing that here.

marcmengel:
So this latest push gets you the split-out mains/cmd.py, mains/submit.py and mains/fetchlog.py with the 
respective mainlines and support code included.   Mostly passes regression tests (except test_condor_unit.py::TestCondorUnit::test_get_schedd_1 and one or two test jobs output tests, which
are unrelated)

shreyb:
I'll re-review, but I think this is good to go.

lib/jobsub_api.py
```diff
@@ -0,0 +1,62 @@
...
+        # put them back
+        sys.stderr = save_err
+        sys.stdout = save_out
+    return output
+
+
+# so clients can easily parse the result strings
+jobsub_submit_re = re.compile(r"Use job id (?P<jobid>[0-9.]+\@[^ ]+) to retrieve")
+
+jobsub_q_re = re.compile(
```
retzkek, line 35: Same as above, but a bit trickier to do "right." Should we add the attributes to the existing `Job` class and return a list of those, or create a new `JobStatus` class specific for this? Simpler API (with some added internal complexity) or easier?

retzkek (changes requested):
We should be careful and deliberate when adding an external API that it's something we can evolve without breaking compatibility, especially knowing how experiment code evolves (or doesn't).

Most APIs will offer high- and low-level interfaces, so I'm ok with offering `jobsub_call()` documented as a low-level interface that accepts arguments with little checking and returns the raw output as a string that is likely to change. It certainly meets the immediate request.

We should also look to offer high-level interfaces that return concrete types, but they don't all have to be (or should be) created now - we can let users drive that by requesting them. A basic `jobsub_submit()` and `jobsub_q()` is fine, and would be mostly wrapping up the already defined regexes.

lib/jobsub_api.py
shreyb: I'd be in favor of adding the attribute to the Job class, and returning a list of those. We wouldn't even have to make up a new enumeration in that case, as the HTCondor python bindings provide those:

https://htcondor.readthedocs.io/en/latest/apis/python-bindings/api/htcondor.html#htcondor.JobStatus

shreyb: @retzkek - given that we plan on addressing these in #609, are you OK with us moving ahead with merging this after my re-review?

marcmengel:
So I went ahead and moved the upper level api to its own pull request, #609 .
That way we can merge this one, and haggle over the upper API in that pull request.
It adds about 400 lines....

retzkek (approved):
LGTM

Issue #596, opened 2025-01-29
Reduce calls to vault for token weakening...


marcmengel:
Currently, when we write a weakened token into our submit file, it is causing condor_vault_storer to fetch and upload a new weakened token
from the vault for every launch.   We should instead remember that we have done this recently for a  given weakened token scope and credd,
and turn off calling condor_vault_storer for later launches to the same credd.

 The easiest may be the following:

- if we perform  a successful submission, we leave a file in /tmp or $HOME/.cache/jobsub_lite or some such indicating the  oauth_handle and credd that we successfully submitted to to remember that condor_vault_storer uploaded a vault token for us for this weakened credential
- when about to submit, if the oauth handle + credd file (above) from a previous launch is in our cache area, and is less than a week old, we set  `_condor_SEC_CREDENTIAL_STORER=/bin/true `and don't even run condor_vault_storer in our submit.

shreyb:
@DrDaveD , @marcmengel , and I talked about this today.  We decided that the best design for this would be to have a flag and corresponding environment variable that would implement the logic as stated above.

So, for example, if I pass this flag to `jobsub_submit`, it would write/check the cache file to see if it was a week old (maybe we need to make that six days), and if it is, we would NOT set `_condor_SEC_CREDENTIAL_STORER` to `/bin/true`.  If the file is newer than that, then as Marc said above, we would set the environment variable.

This should cover both the managed tokens case that a lot of our production users would fall into, but also allow users to set up their own external vault-token renewing scripts if they really wanted to, and not have `jobsub_submit` call `condor_vault_storer` every time.

PR #597, opened 2025-01-30
draft of implementing #596

marcmengel:
So this adds a --managed-token flag to jobsub_submit, which suppresses calling condor_vault_storer when we have
recently called it for this oauth_handle/token scope hash and this schedd.

### DIFF
--- a/lib/condor.py
+++ b/lib/condor.py
@@ -16,11 +16,12 @@
 """ condor related routines """
 from contextlib import contextmanager
 import os
-import sys
+import pathlib
 import random
 import re
 import shutil
 import subprocess
+import sys
 import time
 from typing import Dict, List, Any, Tuple, Optional, Union, Generator
 
@@ -265,10 +266,10 @@ def ran_vault_storer_recently(schedd: str, handle: str, outbase: str) -> bool:
     """See if we left a timestamp file from a previous run of jobsub_submit,
     and it was within the last week minus a bit"""
     try:
-        sv = os.stat(f"{outbase}/.cvs_{schedd}_{handle}")
-        # print(f"saw cvs file, sv.st_mtime is {sv.st_mtime} check: {sv.st_mtime > time.time() - 604500}")
-        # 1 week: 24 * 7 * 60 * 60 = 604800, les 5 minutes (300) -> 604500
-        return sv.st_mtime > time.time() - 604500
+        sv = os.stat(f"{outbase}/.success_vt_store_{schedd}_{handle}")
+        # print(f"saw cvs file, sv.st_mtime is {sv.st_mtime} check: {sv.st_mtime > time.time() - 518400}")
+        # 6 days: 24 * 6 * 60 * 60 = 518400
+        return sv.st_mtime > time.time() - 518400
     except FileNotFoundError:
         return False
 
@@ -276,10 +277,10 @@ def ran_vault_storer_recently(schedd: str, handle: str, outbase: str) -> bool:
 def record_vault_storer_run(schedd: str, handle: str, outbase: str) -> None:
     """touch a file to record when we last ran condor_vault_storer"""
     # recreate/touch the file
-    with open(f"{outbase}/.cvs_{schedd}_{handle}", "w", encoding="utf-8"):
-        pass
-    # print("wrote cvs file...")
+    pathlib.Path(f"{outbase}/.success_vt_store_{schedd}_{handle}").touch(exist_ok=True)
 
+
+NO_OP_STORER = "/bin/true"
 
 # pylint: disable=dangerous-default-value,too-many-locals,too-many-branches,too-many-statements
 @as_span("submit", arg_attrs=["*"])
@@ -325,7 +326,7 @@ def submit(
     if vargs["managed_token"] and ran_vault_storer_recently(
         schedd_name, vargs["oauth_handle"], vargs["outbase"]
     ):
-        _sec_cred_storer_val = "/bin/true"
+        _sec_cred_storer_val = NO_OP_STORER
     else:
         _sec_cred_storer_val = f"{jldir}/bin/condor_vault_storer"
 
@@ -387,7 +388,7 @@ def submit(
             )
             return None
 
-        if vargs["managed_token"] and _sec_cred_storer_val != "/bin/true":
+        if vargs["managed_token"] and _sec_cred_storer_val != NO_OP_STORER:
             record_vault_storer_run(
                 schedd_name, vargs["oauth_handle"], vargs["outbase"]
             )
--- a/lib/condor.py
+++ b/lib/condor.py
@@ -262,22 +262,38 @@ def load_submit_file(filename: str) -> Tuple[Any, Optional[int]]:
     return htcondor.Submit(res), nqueue
 
 
-def ran_vault_storer_recently(schedd: str, handle: str, outbase: str) -> bool:
+def vault_storer_record_filename(schedd: str, handle: str, outbase: str) -> str:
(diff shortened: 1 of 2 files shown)


lib/condor.py
```diff
@@ -260,7 +261,27 @@ def load_submit_file(filename: str) -> Tuple[Any, Optional[int]]:
...
 
 
-# pylint: disable=dangerous-default-value,too-many-locals,too-many-branches
+def ran_vault_storer_recently(schedd: str, handle: str, outbase: str) -> bool:
+    """See if we left a timestamp file from a previous run of jobsub_submit,
+    and it was within the last week minus a bit"""
+    try:
+        sv = os.stat(f"{outbase}/.cvs_{schedd}_{handle}")
+        # print(f"saw cvs file, sv.st_mtime is {sv.st_mtime} check: {sv.st_mtime > time.time() - 604500}")
+        # 1 week: 24 * 7 * 60 * 60 = 604800, les 5 minutes (300) -> 604500
```
DrDaveD, line 270: typo comment `les` -> `less`

lib/get_parser.py
```diff
@@ -479,6 +479,13 @@ def get_parser() -> argparse.ArgumentParser:
         " sequential $JOBSUBJOBSECTION. Scripts may need modification to take"
         " this into account",
     )
+    parser.add_argument(
+        "--managed-token",
+        action="store_const",
+        const=True,
+        default=False,
+        help="Optimize calls to condor_vault_storer, etc.",
```
DrDaveD, line 487: Maybe add `when token is managed` to the help.

lib/condor.py
```diff
@@ -260,7 +261,27 @@ def load_submit_file(filename: str) -> Tuple[Any, Optional[int]]:
...
+        return sv.st_mtime > time.time() - 604500
+    except FileNotFoundError:
+        return False
+
+
+def record_vault_storer_run(schedd: str, handle: str, outbase: str) -> None:
+    """touch a file to record when we last ran condor_vault_storer"""
+    # recreate/touch the file
+    with open(f"{outbase}/.cvs_{schedd}_{handle}", "w", encoding="utf-8"):
```
DrDaveD, line 279: I wonder if anyone would think that `.cvs` refers to the ancient code management system ... maybe `.stored` would avoid confusion.

lib/condor.py
```diff
@@ -266,6 +266,7 @@ def ran_vault_storer_recently(schedd: str, handle: str, outbase: str) -> bool:
     and it was within the last week minus a bit"""
     try:
         sv = os.stat(f"{outbase}/.cvs_{schedd}_{handle}")
+        # print(f"saw cvs file, sv.st_mtime is {sv.st_mtime} check: {sv.st_mtime > time.time() - 604500}")
         # 1 week: 24 * 7 * 60 * 60 = 604800, les 5 minutes (300) -> 604500
         return sv.st_mtime > time.time() - 604500
```
shreyb, line 271: Do we think 5 minutes shy of 7 weeks is a good enough buffer, or should we make it six days?

lib/condor.py
shreyb: Agreed.  I actually was confused about that until I realized it stood for "condor_vault_storer". I'd even say ".stored_token....".

lib/condor.py
```diff
@@ -277,6 +278,7 @@ def record_vault_storer_run(schedd: str, handle: str, outbase: str) -> None:
     # recreate/touch the file
     with open(f"{outbase}/.cvs_{schedd}_{handle}", "w", encoding="utf-8"):
         pass
+    # print("wrote cvs file...")
```
shreyb, line 281: Before we merge this, could you please remove the debug print statements?

lib/condor.py
shreyb: Also, the more idiomatic way of doing this would be:

```python
pathlib.Path(f"{outbase}/.cvs_{schedd}_{handle}").touch(exists_ok=True)
```

That'd be a one-liner (which I still think should be in this function so it's clear to the reader WHY we're touching the file), and you wouldn't need the explanation - it'd be very clear.  Thoughts on using that instead?

Reference:
https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch

lib/get_parser.py
shreyb: Yes - agreed.  I think the help ought to be clear that this is the flag to use when we assume that the vault token will be managed external to jobsub.

lib/condor.py
```diff
@@ -300,7 +321,14 @@ def submit(
     # correct vault token storer script
     # In the condor_vault_storer output, debug gives us more output than verbose,
     # so make that mapping from our verbose/output to condor_vault_storer's
-    _sec_cred_storer_val = f"{jldir}/bin/condor_vault_storer"
+
+    if vargs["managed_token"] and ran_vault_storer_recently(
+        schedd_name, vargs["oauth_handle"], vargs["outbase"]
+    ):
+        _sec_cred_storer_val = "/bin/true"
```
shreyb, line 328: Rather than hardcode `/bin/true` here and below, could we set a global constant called `NO_OP_STORER` or something like that, and set that to `/bin/true`?

Then this line would be very clear that: `_sec_cred_storer_val = NO_OP_STORER`.

lib/condor.py
```diff
@@ -359,6 +387,11 @@ def submit(
             )
             return None
 
+        if vargs["managed_token"] and _sec_cred_storer_val != "/bin/true":
```
shreyb, line 390: See above comment about using a global constant rather than `/bin/true`.

lib/condor.py
marcmengel: Also done.

lib/condor.py
marcmengel: Meh,.  Six is fine, and makes for a simpler comment.

lib/condor.py
marcmengel: Done.

lib/condor.py
marcmengel: Actually, cleaned them up a bit, and made it a verbose print...

lib/condor.py
marcmengel: Also done.

lib/condor.py
```diff
@@ -260,7 +262,27 @@ def load_submit_file(filename: str) -> Tuple[Any, Optional[int]]:
     return htcondor.Submit(res), nqueue
 
 
-# pylint: disable=dangerous-default-value,too-many-locals,too-many-branches
+def ran_vault_storer_recently(schedd: str, handle: str, outbase: str) -> bool:
+    """See if we left a timestamp file from a previous run of jobsub_submit,
+    and it was within the last week minus a bit"""
+    try:
+        sv = os.stat(f"{outbase}/.success_vt_store_{schedd}_{handle}")
```
shreyb: One more thing I just thought of.  Since we use this path (`f"{outbase}/.success_vt_store_{schedd}_{handle}"`) in multiple places, what would you think of having a function that returns the filename based on the outbase, schedd, and handle?

Something like:

```python
def some_name(schedd, handle, outbase):
  return f"{outbase}/.success_vt_store_{schedd}_{handle}"
```

And then call that function in `ran_vault_storer_recently` and `record_vault_storer_run`? 

The upside would be that if we ever changed the naming convention, we'd only make the change in one place.  The downside of course is it slightly complicates the functions here (e.g. `os.stat(some_name(schedd, handle, outbase))`).

shreyb: Did you read my mind?!?

shreyb (approved):
I'm good with this.  @DrDaveD  - any further comments?

PR #598, opened 2025-02-04
Added env JOBSUB_MANAGED_TOKEN to control --managed-token flag default

shreyb:
This PR adds an environment variable, `JOBSUB_MANAGED_TOKEN`, corresponding to the new `--managed-token` flag, that allows users to set that flag in the environment

```
# Default behavior
$  bin/jobsub_submit -G fermilab -n --debug file:///usr/bin/true
...
varg: {...  'managed_token': False, 'memory': '2GB', ...}
...

# With flag
$  bin/jobsub_submit -G fermilab -n --debug --managed-token file:///usr/bin/true | grep managed_token
...
varg: {... 'managed_token': True, 'memory': '2GB', ...}
...

# With env
$ JOBSUB_MANAGED_TOKEN=1 bin/jobsub_submit -G fermilab -n --debug file:///usr/bin/true
...
varg: {...  managed_token': True, 'memory': '2GB', ...}
...
```

There are also a couple of tests that check this.

### DIFF
--- a/lib/get_parser.py
+++ b/lib/get_parser.py
@@ -483,8 +483,10 @@ def get_parser() -> argparse.ArgumentParser:
         "--managed-token",
         action="store_const",
         const=True,
-        default=False,
-        help="Optimize calls to condor_vault_storer, etc. when using managed tokens",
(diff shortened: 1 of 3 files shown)


marcmengel (approved):
I like it!

PR #601, opened 2025-02-05
Add a "did you mean" clause for invalid but close to right auth methods

### LINKED ISSUE #595: In the case of auth methods, we should have a "did you mean..." error message in case of invalid auth method
Nova had a case recently where they set all user auth methods to tokens using the ENV: `JOBSUB_AUTH_METHODS=tokens`.  That's a typo - the value should be `token`.  But jobsub just complains about an invalid auth method, rather than give a helpful message.

For some finite set of wrong auth methods, like "tokens, proxies, Token, Proxy", etc., we could print a warning message telling the user what the correct strings are.

It's also remarkably easy for a user to use `JOBSUB_AUTH_METHOD` instead of `JOBSUB_AUTH_METHODS`.  We should probably support both ENVs and have them do the same thing.

shreyb:
Closes #595 

This just adds a `difflib` check that drives a "Did you mean..." error message.  This PR includes accompanying tests.

For example:

```
# Current default

$ bin/jobsub_submit -G fermilab --auth-methods token,proxy -n file:///bin/true                             
Submission files are in: /home/sbhat/.cache/jobsub_lite/js_2025_02_05_174257_7cbc693c-e519-483a-914d-1a381e3b796b


# Good case
$ bin/jobsub_submit -G fermilab --auth-methods token -n file:///bin/true
Storing bearer token in /tmp/bt_token_fermilab_Analysis_10610
Submission files are in: /home/sbhat/.cache/jobsub_lite/js_2025_02_05_174223_fc719f07-5266-4aae-a97b-25cfb2b8daf8

# tokens instead of token
$ bin/jobsub_submit -G fermilab --auth-methods tokens -n file:///bin/true                                  


Error: ValueError: The jobsub_lite infrastructure requires that the following authorization methods be present: ['token']. You provided 'tokens' - did you mean 'token'?

# proxies instead of proxy
$ bin/jobsub_submit -G fermilab --auth-methods token,proxies -n file:///bin/true                           


Error: ValueError: Invalid auth method proxies.  Supported auth methods are ['token', 'proxy']. You provided 'proxies' - did you mean 'proxy'?


# tokens instead of token, but token is required, so it'll fail
$ bin/jobsub_submit -G fermilab --auth-methods tokens,proxy -n file:///bin/true                            


Error: ValueError: The jobsub_lite infrastructure requires that the following authorization methods be present: ['token']. You provided 'tokens' - did you mean 'token'?
```

### DIFF
--- a/lib/get_parser.py
+++ b/lib/get_parser.py
@@ -20,7 +20,7 @@ import difflib
 import os
 import re
 import sys
-from typing import Union, Any
+from typing import Union, Any, List
 
 from condor import get_schedd_names
 from creds import SUPPORTED_AUTH_METHODS, REQUIRED_AUTH_METHODS
@@ -152,7 +152,7 @@ class CheckIfValidAuthMethod(argparse.Action):
         setattr(namespace, self.dest, ",".join(check_values))
 
     @staticmethod
-    def __get_msg_from_close_val(value: str, valid_values: list[str]) -> str:
+    def __get_msg_from_close_val(value: str, valid_values: List[str]) -> str:
         """If our value is close to a valid value, return a message asking if the user meant
         the valid value.  Otherwise, return an empty string"""
         did_you_mean_str = " You provided '{value}' - did you mean '{close_match}'?"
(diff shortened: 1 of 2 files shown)


vitodb (approved):
This looks nice!

vitodb (approved):
This looks good!

PR #607, opened 2025-02-28
using Spack to find ifdh, and adding logging to sambegin.sh, samend.sh

marcmengel:
Have sambegin.sh and samend.sh find ifdhc  via Spack, as the simple.sh wrapper does. 

Also add some ifdh log calls to sambegin.sh and samend.sh to facilitate debugging.


shreyb:
Looks good to me.  Merging, and it will go out with 1.12.

shreyb (commented):
Forgot to mark approval here.  Oops.

PR #609, opened 2025-03-07
Upper api

marcmengel:
This is adding an upper level api to the big_jobsub_api branch / pull #594, which has 2 main calls,
and an extenstion of the Job class, SubmittedJob.

The two bare entry points are 
- submit( group="...", executable="..." ...)  which has a plethora of jobsub_submit arguments and returns a SubmittedJob object for the newly submitted job and
- q(group="...", user=""..."...) which returns a list of SubmittedJob objects, one for each entry returned by jobsub_q with the mapped argumetns.  It may need more arguments mapped to meet peoples neesds.

Job objects  (existing in condor.py) have members
-  id for jobids
- cluster
- proc
- schedd

SubmittedJob objects add
- pool
- group
- role
- auth_methods
- submit_output (onlyt if the result of submit())
- owner  (if jobsub_q output or q() methods called)
- submitted (datetime.datetime) ( "" )
- runtime (datetime.timedelta) ( "" )
- status (htcondor.JobStatus) (  "" )
- prio
- size
- command

as well as methods:
- hold()
- release()
- rm()
- fetchlog( destdir, condor=False) 
- q()  (update owner, status, etc. with jobsub_q)
- q_long() (...and return --long info as dictionary)
- q_analyze() (return jobsub_q --better-analyze output)
- wait() (run q() periodically until status COMPLETED, HELD, or REMOVED.)

Any of those SubmittedJob  objects have the ability to manage the jobs with the methods mentioned above.


marcmengel:
Recent playing with new api interactively:
[demo.txt](https://github.com/user-attachments/files/19133278/demo.txt)

lib/jobsub_api.py
```diff
@@ -74,3 +90,519 @@ def jobsub_call(argv: List[str], return_output: bool = False) -> Optional[str]:
         print(f"Excepion in jobsub_call({argv})")
         raise
     return res
+
+
+# =-=-=-=-=-=-=-=-=-=-=-=-=-=-=
+
+
+def optfix(s: Optional[str]) -> str:
```
shreyb, line 98: Rather than have this function wrapping every `jobsub_call` invocation, perhaps it's better to have `jobsub_call` only return a string.  Looking at that function, it looks like the only reason it needs to be `Optional[str]` is because you initialize `res` to `None`.  Maybe if that's `""`, then you can always return a string, and then this function won't be necessary.

lib/jobsub_api.py
```diff
@@ -74,3 +90,519 @@ def jobsub_call(argv: List[str], return_output: bool = False) -> Optional[str]:
...
+            args.append(self.auth_methods)
+        if self.role:
+            args.append("--role")
+            args.append(self.role)
+        args.append(self.id)
+
+    def hold(self, verbose: int = 0) -> str:
+        """Hold this job with jobsub_hold"""
+        args = ["jobsub_hold"]
+        self._update_args(verbose, args)
```
shreyb, line 170: Since these last three lines are run for hold, release, and rm, could they be combined into a single function, which gets called from each of those funcs?  Something like:

```python
def some_better_name(self, command, verbose):
    args = [command]
    self._update_args(verbose, args)
    rs = optfix(jobsub_call(args, True))
    return rs

def hold(self, verbose):
    return self.some_better_name("jobsub_hold", verbose)

def release(self, verbose):
    return self.some_better_name("jobsub_release", verbose)

def rm(self, verbose):
    return self.some_better_name("jobsub_rm", verbose)
```

marcmengel: So that was the Idea of the _update_args routine already, so I collapsed it into one called _cmd() that takes
the initial args with the base command and  base options, and moved the optfix(jobsub_call...) up there.

lib/jobsub_api.py
```diff
@@ -70,7 +86,506 @@ def jobsub_call(argv: List[str], return_output: bool = False) -> Optional[str]:
...
+    # pylint: disable=invalid-name
+    def rm(self, verbose: int = 0) -> str:
+        """Remove this job with jobsub_rm"""
+        return self._cmd(verbose, ["jobsub_rm"])
+
+    def q(self, verbose: int = 0) -> None:
+        """run 'jobsub_q' on this job and update values, status"""
+        rs = self._cmd(verbose, ["jobsub_q"])
+        lines = rs.split("\n")
+        if len(lines) == 2 and self.status is not None:
```
shreyb, line 185: I worry about specifying the lines count to see if the `jobsub_q` output.  For example in this case, I think the current check breaks:

```
[sbhat@host~]$ jobsub_q -G uboone --jobid=1@schedd1.fnal.gov
Attempting to get token from https://vaultserver.fnal.gov:8200 ... succeeded
Storing bearer token in /tmp/bt_token_uboone_Analysis_UID
JOBSUBJOBID                             OWNER           SUBMITTED     RUNTIME   ST PRIO   SIZE  COMMAND
0 total; 0 completed, 0 removed, 0 idle, 0 running, 0 held, 0 suspended
```

By my count, that's 4 lines, but it could apply to a job that we previously saw, but has now left the queue. Maybe this check needs to be more specific, perhaps something that looks for the standard `jobsub_q` header, in the lines, and then sees if the next element of lines has the jobid in it?  Just thinking "out loud" here, I think that could look something like:

```python
def has_q_output(self, q_lines) -> bool:
    header_idx = 0
    for i, line in enumerate(lines):
        if re.search("^JOBSUBJOBID\tOWNER.+$", line):
            header_idx = i
            break
    else:
        return False
    
    try:
        return lines[header_idx + 1].find(self.jobid) != -1:
    except IndexError:
        return False
```

There may be a more intelligent way than this example to check the `jobsub_q` output - but either way, I think we should make the check stronger.

lib/jobsub_api.py
```diff
@@ -70,7 +86,506 @@ def jobsub_call(argv: List[str], return_output: bool = False) -> Optional[str]:
...
+            # we saw it previously, and now it is not showing up..
+            self.status = JobStatus.COMPLETED
+            return
+        if len(lines) > 1:
```
shreyb, line 189: Now reading this part, if we implement an output checker like above, the logic of this whole method could be something along the lines of:

```python
if not self.has_q_output(lines):
   if self.status is not None:
      self.status = JobStatus.COMPLETED
      return
   raise RuntimeError(....)

# This is the happy path
line = rs.split("\n")[1]
...
return
```

I think that helps with readability as well - what do you think?

lib/jobsub_api.py
```diff
@@ -70,7 +86,506 @@ def jobsub_call(argv: List[str], return_output: bool = False) -> Optional[str]:
...
+        lines = rs.split("\n")
+        res = {}
+        for line in lines:
+            if not line.find(" = ") > 0:
+                continue
+            k, v = line.split(" = ", 1)
+            if v[0] == '"':
+                v = v.strip('"')
+            res[k] = v
+        if len(lines) == 1 and self.status is not None:
```
shreyb, line 218: VERY nitpicky here, but I think the intent of this would be clearer if rather than checking `len(lines)`, you check to see if `rs == ""`, or even `if not rs`. This case applies to a blank result from rs, if I understand correctly, and one of those ways would communicate that clearer.

The other issue is that len(lines) COULD be 1 and the result of rs non-blank and still if the condor devs in the future forget to put line breaks in. To illustrate the point

```
>>> lines = "".split("\n")
>>> lines, len(lines)
([''], 1)
>>> lines = "blahblahoopsnolineendingkey=value".split("\n")
>>> lines, len(lines)
(['blahblahoopsnolineendingkey=value'], 1)
```

marcmengel: Well, here I'm trying to handle the case that jobs just stop showing up when they complete around here; so you do
jobsub_q and it comes up empty.   When you're doing plain jobsub_q , you still get the headers, so length(lines)==2, (["JOBSUBJOBID ...", ""])l;  but for jobsub_q --long we don't get the header so length(lines)==1... So I was trying to have parallel tests in the two cases.   But you're right, at least a comment here about what the len(lines) == n means would be appropriate.

marcmengel: Added a little dissertation on detecting the emtpy result, and why we're marking it "COMPLETED", in comments.

shreyb: I understand where you're coming from, but I'm still not too keen on the check counting lines, if there's a more reliable way.  How about some internal function that checks both cases?  Like a 

```python
def _jobsub_q_output_nonempty(stdout: str, long: bool) -> bool:
    # logic if long:
    if long:
        return stdout != "" 
   # logic if not long
   stuff()
```

Then you could have the same function called in both cases, so you'd satisfy having parallel checks.

retzkek:
We should also give some thought to how this API will be documented for users. 

Conveniently that's the topic of tomorrow's CSAID roadmap meeting.

shreyb:
I'm sorry I missed the context of your comment in the meeting earlier today, Kevin.  Yes, we should have this be a part of #612 as well.  It's kind of in both items' scopes.

lib/jobsub_api.py
marcmengel: Changed per shreyb's suggestion.   Initialized to "" , and got rid of optfix() and Optional import. [upper_api 44fd798]

lib/jobsub_api.py
marcmengel: So upon reflection, the simpler, better check is if `self.jobid` is in the whole output string.
```
   def q(self, verbose: int = 0) -> None:
        """run 'jobsub_q' on this job and update values, status"""
        rs = self._cmd(verbose, ["jobsub_q"])
        if rs.find(self.jobid) < 0:
             ...
```

lib/jobsub_api.py
marcmengel: Here also switched to looking for self.id in the output.

lib/jobsub_api.py
```diff
@@ -70,7 +86,506 @@ def jobsub_call(argv: List[str], return_output: bool = False) -> Optional[str]:
...
+        """run 'jobsub_q --long' on this job and update values, status,
+        and return the dictionary of name/value pairs from the ouput"""
+        rs = self._cmd(verbose, ["jobsub_q", "--long"])
+        lines = rs.split("\n")
+        res = {}
+        for line in lines:
+            if not line.find(" = ") > 0:
+                continue
+            k, v = line.split(" = ", 1)
+            if v[0] == '"':
```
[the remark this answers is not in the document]
shreyb: Following up here - Is there any reason you're doing the `v[0]` check?  You can just say `v = v.strip('"')`, and if there's no `"`, it's a no-op.

lib/jobsub_api.py
shreyb: Sorry - I must be missing something.  Where is the `self.id` check in the output?

tests/test_api.py
```diff
@@ -42,3 +44,31 @@ def test_api_demo():
...
+        print(f"submitted job: {job1.id}")
+        print(f"submit output:\n=-=-=-=-=\n {job1.submit_out}")
+        print(f"\n=-=-=-=-=\n")
+        qjobs = q(job1.id, group=group)
+        for qjob in qjobs:
+            print(f"saw job {qjob.id} status {str(qjob.status)} ")
+        rs = job1.fetchlog(destdir="/tmp/test_fetch", verbose=1)
+        print(f"fetchlog says: {rs}")
+        sb = os.stat("/tmp/test_fetch/lookaround.sh")
+        assert time.time() - sb.st_mtime < 5
```
shreyb, line 68: I believe this check is to make sure we properly generated a new file in `/tmp/test_fetch` during the fetchlog.  I think a cleaner way to do it would be to use the `tmp_path` fixture from pytest to create your test directory, and set that to be the `destdir` argument a few lines above in `rs = job1.fetchlog(...`. Then, you wouldn't have to try to remove that directory either on line 69: pytest would take care of it for you.

This would also make the test cleaner because you wouldn't have to check that the generated `/tmp/test_fetch/lookaround.sh` was a new file - you'd just have to make sure it exists in the directory returned by `tmp_path`.

I've got a lot of experience using `tmp_path` - would you like me to modify this test as I've written in this comment, or did you want to play around with it first (I know you're SUPER busy).

tests/test_api.py
```diff
@@ -42,3 +44,31 @@ def test_api_demo():
...
+        os.system("rm -rf /tmp/test_fetch")
+        data = job1.q_long()
+        print(f"after update, status: {str(job1.status)}")
+        assert "ClusterId" in data
+    except RuntimeError:
```
shreyb, line 73: I'm not sure this is necessary: If there's a `RuntimeError` while running the test, it'll just fail, which is what we want.

lib/jobsub_api.py
shreyb: Ah, I see.  You implemented the `self.id` check above in `q()`, but not here.  I like that idea to do it in both places - could you do that here as well?

lib/jobsub_api.py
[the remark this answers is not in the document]
marcmengel: Again, forgot to push.

tests/test_api.py
marcmengel: If you could add it quickly, that would be awesome.

tests/test_api.py
marcmengel: I think I had some debug printing in there at one point...

marcmengel: I think the real reason was, I wanted this to be a steal-able example of using the API, and the things that
in a shell script exit non-zero raise RunTimeError in the API.   So I think I'll re-add the print, and leave it in
for just that reason.

tests/test_api.py
shreyb: Sure thing - I'll try to do it today, but I can definitely get that in before the end of this week.

shreyb:
I made two changes here:  

1) I updated the fancy_api_demo test to use the pytest `tmp_path` fixture, as we discussed
2) In commit [270e35b](https://github.com/fermitools/jobsub_lite/pull/609/commits/270e35b50af8e9bff89439d454a1f365f1ba543f), I fixed a long-standing bug in jobsub_fetchlog (using the condor sandbox) where we referred to a variable when it might not have been assigned.  That fix is in `lib/mains/fetchlog.py -->fetch_from_condor()`, so please feel free to make sure it makes sense.

tests/test_api.py
shreyb: Done.  See PR conversation

shreyb:
Are there any other requests/changes we want to make to this API before we ask @marcmengel to formally mark the PR for review?  @retzkek ?

retzkek:
> Are there any other requests/changes we want to make to this API before we ask @marcmengel to formally mark the PR for review? @retzkek ?

I do not, go ahead and remove draft @marcmengel.

lib/jobsub_api.py
```diff
@@ -70,7 +86,528 @@ def jobsub_call(argv: List[str], return_output: bool = False) -> Optional[str]:
...
+    "tar_file_name": "--tar_file_name",
+    "timeout": "--timeout",
+    "t": "-t",
+    "verbose": "--verbose",
+}
+
+
+# pylint: disable=dangerous-default-value, too-many-branches, too-many-arguments, invalid-name
+def submit(
+    executable: str,
```
shreyb, line 341: Something I just realized - we require the `-G/--group` flag for `jobsub_submit`.  We should also require it here in the function signature, with a default being `os.environ.get("GROUP", None)`.

Sorry I missed this earlier.

lib/jobsub_api.py
```diff
@@ -70,7 +86,528 @@ def jobsub_call(argv: List[str], return_output: bool = False) -> Optional[str]:
         with output_saver(return_output) as output:
             func(argv)
             res = output.getvalue()
-    except:
-        print(f"Excepion in jobsub_call({argv})")
-        raise
+    except Exception as e:
+        raise RuntimeError(f"Excepion in jobsub_call({argv})") from e
```
shreyb, line 90: Typo here - "Excepion" should be "Exception".

man/jobsub_api.md
```diff
@@ -0,0 +1,172 @@
...
+* schedd -- schedd job was submitted to
+
+SubmittedJob objects add
+* pool
+* group
+* role
+* auth_methods
+* submit_output (onlyt if the result of submit())
+* owner  (if jobsub_q output or q() methods called)
+* submitted (datetime.datetime) ( "" )
```
shreyb, line 156: This and the next two lines are wrong.  The default values, according to jobsub_api.py, are `None`, not `""`.

shreyb (commented):
I found a couple of issues while going through the documentation. The big issue, though, is that I think we should add `group` as an explicit argument to both `submit` and `q`, with the default being the environment setting.

lib/jobsub_api.py
```diff
@@ -70,7 +91,530 @@ def jobsub_call(argv: List[str], return_output: bool = False) -> Optional[str]:
...
+    "role",
+    "subgroup",
+    "constraint",
+    "name",
+    "user",
+]
+
+
+def q(
+    *jobids: str, devserver: bool = False, verbose: int = 0, **kwargs: str
```
shreyb, line 564: Same comment here as with `submit` - that we should have `group` be an explicit argument here, to communicate that we expect it to be provided somehow.

shreyb (changes requested):

man/jobsub_api.md
marcmengel: Actually that was supposed to be "ditto marks" , meaning repeat the " (if jobsub_q output or q() methods called)"...  Sigh.

retzkek (approved):

lib/jobsub_api.py
shreyb: This one still is out there.

marcmengel: Speling fixed.

PR #619, opened 2025-03-21
Fixed unit test after --maxConcurrent type restriction

shreyb:
This unit test was assuming that maxConcurrent could be anything.  Now that it's an int, the test broke.  This PR fixes it.

Also, we had forgotten to coerce maxConcurrent to 0 if we were not using a DAG in certain cases.


marcmengel (approved):
LGTM

PR #620, opened 2025-03-31
Fix IFDH_VERSION behavior in job wrapper.

marcmengel:
It *should* activate the spack environment ending in $IFDH_VERSION, but instead 
the `echo &&` breaks desired behavior, and both spack environment activate  commands run 
and you end up with the last one...

### DIFF
--- a/lib/mains/cmd.py
+++ b/lib/mains/cmd.py
@@ -248,7 +248,7 @@ def jobsub_cmd_args(arglist: argparse.Namespace, passthru: List[str]) -> None:
                 # pipe our remaining output through sort (by date) and jobsub_totals next to us
                 # ... if we're not collecting output for the API.
                 jobsub_totals_path = os.path.join(
-                    os.path.dirname(__file__), "../bin/jobsub_totals"
+                    os.path.dirname(__file__), "../../bin/jobsub_totals"
                 )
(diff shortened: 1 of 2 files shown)


marcmengel:
This should definitely be included in whatever release we have upcoming soonest, it prevents people from properly testing new ifdhc releases in their jobs.

vitodb (approved):
this looks good!

shreyb:
> This should definitely be included in whatever release we have upcoming soonest, it prevents people from properly testing new ifdhc releases in their jobs.

The only issue there is that the huge code refactor is taken into consideration with this PR.  I'm cutting the 1.11 release now, to go out tomorrow.  Could we replicate the changes in this PR for the old (currently-released) version of the code, and then release that next week as 1.11.1? I agree that it'd be nice to have this go out ASAP, and if I read this PR correctly, it's JUST the `simple.sh` that would need to change.

Thoughts?

marcmengel:
Yes.  I'll make a separate PR against... 1.11-turn-off-proxies(?) Does that sound like the right branch?

marcmengel:
It's in: #622

marcmengel:
I note that this also has a second small fix for the jobsub_api stuff, the finding of the totals script for jobsub_q which adds the cross-schedd job totatls needs an extra subdirectory dealt wtih.

shreyb:
Since #624 is merged, I'll merge this into `master`.

PR #622, opened 2025-04-07
Fix IFDH_VERSION on  1 11 branch

marcmengel:
This is pull #620, but against the 1.11-turn-off-proxies branch.


marcmengel:
Made against wrong branch. Cancelling.

PR #623, opened 2025-04-07
fix IFDH_VERSION against 1.11 release

marcmengel:
Once more, with feeling.


marcmengel:
Sigh.  I rebased the branches under this, and now its listing all sorts of unwanted changes... Sigh.

Issue #626, opened 2025-04-08
Fix spec file to work with new API library structure


shreyb:
Our spec file causing the CI tests to fail.  While we would have fixed this with the 1.12 release anyway, we might as well fix it before that.

```
RPM build errors:
   /opt/jobsub_lite/lib/jobsub_api.py
    File not found: /github/home/rpmbuild/BUILDROOT/jobsub_lite-1.10-1.el9.x86_64/opt/jobsub_lite/lib/__pycache__/*
    File not found: /github/home/rpmbuild/BUILDROOT/jobsub_lite-1.10-1.el9.x86_64/opt/jobsub_lite/*/*.py?
    Installed (but unpackaged) file(s) found:
   /opt/jobsub_lite/lib/jobsub_api.py
make: *** [Makefile:37: rpm] Error 1
```

PR #630, opened 2025-04-25
Updated PAT to fine-grained token for better security

shreyb:
As the title says.  The token used to checkout `jobsub_lite_config` now can ONLY do that.


shreyb:
Though the affected check won't run here, I pushed a copy of this branch to the `fermitools/jobsub_lite` repo, and ran the action manually. As expected, the action successfully checked out `jobsub_lite_config` with the new token, and failed when running Make (see #626 ). So this change itself works, and once merged to `master`, should fail at the same spot as my test.

marcmengel (approved):

PR #637, opened 2025-06-04
Update config repo commit to a40a2

### LINKED ISSUE #611: Change spec file to include new files
The API change from #594 and #609 has added a lot of new files. We need to update the spec file in jobsub_lite_config to reflect that, else our github build (and eventual RPM creation) will fail.

shreyb:
Closes #611 

This adds the jobsub_api library files to the spec file that is used in the RPM as well as in our build tests.


shreyb:
I forgot to note in jobsub_lite_config PR 30 - I tested creating the RPM with the new spec from that PR file using the `Makefile` in this repository.

marcmengel (approved):
Looks good

PR #645, opened 2025-08-06
Trace jobsub api

marcmengel:
This adds @as_span() decorators to API calls so we can tell what was called via the API vs otherwise.


shreyb (commented):
LGTM except for the one question I raised.

shreyb (approved):
LGTM

PR #657, opened 2026-01-05
Make sure we check for executable's existence before submitting job

### LINKED ISSUE #631: Upgrade executable verification action to check for file existence (file://)
https://github.com/fermitools/jobsub_lite/blob/90d5385659599b26609bb1b7d5c38ca691f5a190/lib/get_parser.py#L35

We noted here that this check should be beefed up to check for the existence of the user-given executable.  Not having that check just bit us (see ServiceNow ticket INC000001194303), so we should implement the check.

shreyb:
Closes #631 

This PR upgrades the `executable` argument in the jobsub `get_parser` parser to use a custom `Action` that does two things:

1. Check that the executable argument is given with the correct URI (`file://`)
2. Check that the argument value for the executable actually exists.

There are added accompanying unit tests, as well as a couple of updates to old tests so that they work with this new restriction.

All unit tests pass, and further testing information is available in comments of #631.

### DIFF
--- a/lib/get_parser.py
+++ b/lib/get_parser.py
@@ -29,16 +29,6 @@ import pool
 from skip_checks import SupportedSkipChecks
 
 
-def verify_executable_starts_with_file_colon(s: str) -> str:
-    """routine to give argparse to verify the executable parameter,
-    which is supposed to be given as a file:///path URL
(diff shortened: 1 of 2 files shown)


vitodb (approved):
This looks good

PR #664, opened 2026-04-22
Added support for using htcondor2 and classad2 libraries

### LINKED ISSUE #663: Add support for condor 25
The biggest change to condor 25 so far is the change in the module name of the condor python bindings - from `htcondor` and `classad` to `htcondor2` and `classad2`, respectively.

We should see if we can support both - given the limited usage of python bindings in jobsub, this should hopefully be pretty straightforward.

Here's what happens right now if we use jobsub 1.12 and condor 25:

```
$ jobsub_submit -G fermilab file:///bin/bash
  File "/opt/jobsub_lite/bin/jobsub_submit", line 41, in <module>
    from mains.submit import jobsub_submit_main, VERBOSE
  File "/opt/jobsub_lite/lib/mains/__init__.py", line 2, in <module>
    from .cmd import jobsub_cmd_parser, jobsub_cmd_main, jobsub_cmd_args
  File "/opt/jobsub_lite/lib/mains/cmd.py", line 16, in <module>
    import condor
  File "/opt/jobsub_lite/lib/condor.py", line 29, in <module>
    import classad  # type: ignore
ModuleNotFoundError: No module named 'classad'
```

This is because of the renaming of the module.  Hopefully the solution is as simple as what the migration document [here](https://htcondor.readthedocs.io/en/lts/apis/python-bindings/api/version2/migration-guide.html) says:  

```python
import htcondor2 as htcondor
import classad2 as classad
```

And to make it support both condor ≤24 AND condor 25:

```python
try:
  import htcondor2 as htcondor
except ModuleNotFoundError: # condor 24 or lower
  import htcondor
```

in every file that imports `htcondor` or `classad` currently. But we will want to check.  The following are the files that currently import `htcondor` or `classad` (using git grep for `import htcondor`, `import classad`, `from htcondor`, or `from classad`):

## import `htcondor`
1. bin/condor_submit
2. lib/condor.py
3. lib/cred_token.py
4. lib/mains/fetchlog.py
5. lib/jobsub_api.py

## import `classad`
1. lib/condor.py
2. lib/utils.py

shreyb:
This PR adds conditional imports of the `htcondor2` and `classad2` Condor Python bindings.  We want to still support both `htcondor` and `classad`, since some client machines are running Condor 23, which does not have the `2` versions of those libraries.  

The plan is that in the upcoming release (1.13), we will support both `htcondor`/`classad` and the `2` versions of those, and then drop support for `htcondor`/`classad` in the 1.14 release.

This code has been exhaustively tested on the following condor combinations (each combination below is client --> schedd):

1. Condor 23 --> Condor 25 (Current FNAL production)
2. Condor 24 --> Condor 25
3. Condor 25 --> Condor 25 (Near future FNAL production)
4. Condor 25 --> Condor 24 (`condor_vault_storer` FAILS, but we never intend to use this combination, and it is probably independent of this change.)

Closes #663.

### DIFF
--- a/bin/condor_submit
+++ b/bin/condor_submit
@@ -17,7 +17,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-""" wrapper for condor submit """
+"""wrapper for condor submit"""
 # pylint: disable=wrong-import-position,wrong-import-order,import-error
 import sys
 import os
@@ -28,7 +28,10 @@ if os.environ.get("LD_LIBRARY_PATH", ""):
     del os.environ["LD_LIBRARY_PATH"]
     os.execv(sys.argv[0], sys.argv)
 
-import htcondor
+try:
+    import htcondor2 as htcondor  # Condor 25 and later
+except ImportError:
+    import htcondor  # Condor 24 and earlier
 
 #
 # we are in prefix/bin/jobsub_submit, so find our prefix
--- a/lib/condor.py
+++ b/lib/condor.py
@@ -26,8 +26,13 @@ import time
 from typing import Dict, List, Any, Tuple, Optional, Union, Generator
 
 # pylint: disable=import-error
-import classad  # type: ignore
-import htcondor  # type: ignore
+try:  # Condor 25 and later
+    import classad2 as classad  # type: ignore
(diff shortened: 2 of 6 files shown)


vitodb (approved):
This looks good!

PR #665, opened 2026-04-27
pre-commit.ci fixes 1.13

shreyb:
This fixes the various pre-commit.ci issues seen in PR #664.

The fixes were:
- Upgrade mypy to v1.10.0 so it uses the python `ast` library 
- Upgrade black to 22.8.0 so it uses the updated async calls in the python stdlib.


vitodb (approved):
This looks good!

PR #670, opened 2026-05-12
Fixed case where external callers (like POMS) run fake_ifdh.py getToken

shreyb:
As @marcmengel, @vitodb, and I discussed.

1. In `lib/creds.py:getRole`, change argument name from `role_override` to `role`.
2. In `lib/fake_ifdh.py`, make sure that if we're giving role to a random command-called function, we specify the `role` keyword. Also, if we have errors, exit with a non-zero exit code.
3. Added tests to check a few cases (incomplete, but covers the issue we found in testing).


marcmengel (approved):
LGTM

Issue #599, opened 2025-02-05
A number of custom action `TypeError`s should actually be `ValueError`s


shreyb:
In `lib/get_parser.py`, there are a number of custom actions for which invalid user input results in a `TypeError`.  According to the python [documentation](https://docs.python.org/3.9/library/exceptions.html#TypeError): 

> Passing arguments of the wrong type (e.g. passing a [list](https://docs.python.org/3.6/library/stdtypes.html#list) when an 
> [int](https://docs.python.org/3.6/library/functions.html#int) is expected) should result in a 
> [TypeError](https://docs.python.org/3.6/library/exceptions.html#TypeError), but passing arguments with the wrong value (e.g. a number
> outside expected boundaries) should result in a [ValueError](https://docs.python.org/3.6/library/exceptions.html#ValueError).

We should convert all of those custom action errors to `ValueError`, and make sure that the `get_parser` tests are updated as well.

PR #671, opened 2026-07-07
Raise ValueError for invalid custom-action argument values (#599)

arpitjain099:
This addresses #599. The custom argparse actions in `lib/get_parser.py` were raising TypeError when a user passed a bad *value* to `--skip-check`, `--schedd-for-testing`, or the executable argument, but as the issue notes, the Python docs reserve TypeError for wrong types and ValueError for wrong values. I switched those three raises to ValueError and updated the matching expectations in `tests/test_get_parser_unit.py`.

I ran `pytest tests/test_get_parser_unit.py` for the affected cases: the --skip-check and executable checks are pure and pass with the change (and fail if I revert the raise back to TypeError, so the tests are actually exercising it). The invalid-schedd case goes through `get_schedd_names`, so I couldn't fully exercise it without a live collector, but the assertion change matches the other two. Thanks for taking a look.

### DIFF
--- a/lib/get_parser.py
+++ b/lib/get_parser.py
@@ -73,7 +73,7 @@ class VerifyAndAddSkipCheck(argparse.Action):
     ) -> None:
         _supported_checks = SupportedSkipChecks.get_all_checks()
         if values not in _supported_checks:
-            raise TypeError(
+            raise ValueError(
                 f'Invalid argument to flag --skip-check: "{values}". Value must '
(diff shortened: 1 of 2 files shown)


shreyb:
LGTM.  I checked out this code and ran unit tests against it, including integration tests, and everything works.

shreyb:
Thank you for the contribution, @arpitjain099!

